PR #4, opened 2018-05-29
Add nginx proxy server

Bootjack:
This serves both the tileserver and the rails app from the same host:port so that the client can just load tile data from `/tiles`.


luandro:
@codex review

PR #32, opened 2018-06-09
WIP: MySQL -> Postgres [do not merge]

seanlinsley:
this builds off of https://github.com/bitnami/bitnami-docker-rails/pull/75
in order to make use of Postgres GIS via https://github.com/rgeo/activerecord-postgis-adapter

unfortunately Rails isn't able to connect to the database... it seems to be configured correctly though

### DIFF
--- a/rails/config/database.yml
+++ b/rails/config/database.yml
@@ -10,12 +10,13 @@
 #   https://dev.mysql.com/doc/refman/5.7/en/password-hashing.html
 #
 default: &default
-  adapter: mysql2
+  adapter: postgresql
   encoding: utf8
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   username: root
   password:
-  host: mariadb
+  host: postgres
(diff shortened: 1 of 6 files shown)


DarthHater:
What is the error that is returned when trying to connect to Postgres? When running postgres in a docker-compose environment, I think you need to pass ?sslmode=disable in the connection string.

rails/config/database.yml
```diff
@@ -10,12 +10,13 @@
 #   https://dev.mysql.com/doc/refman/5.7/en/password-hashing.html
 #
 default: &default
-  adapter: mysql2
+  adapter: postgresql
   encoding: utf8
   pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
   username: root
```
DarthHater: Pretty sure you need to provide a username/password for postgres. The documentation for the image you are using shows one being sent via -e POSTGRES_PASSWORD in the run command.

PR #42, opened 2018-06-09
Add Speaker and Story relations

lizthree:
1:many - speaker:story
1:1 - story:speaker


seanlinsley:
the site failed to load b/c the fake stories being used to render the list view in the card were passing a string to the now-removed story attribute. I pushed a commit to fix that

PR #75, opened 2018-07-01
[73] Add webpacker with React support

Bootjack:
I got it! I think I got it!

ithinkigotit


Bootjack:
I'm considering this finished enough to merge. This is a straightforward port from our view partials to React components. There's certainly more we could do to improve the composition of our component hierarchy and we need to establish a common pattern for updating story/point/media data based on user interaction, beyond the initial data included on pageload.

Issue #76, opened 2018-07-05
Update the introduction text "What is a Terrastory?"


Bootjack:
Use the following text:

Terrastories are audiovisual recordings of place-based storytelling. This offline-compatible application enables local communities to locate and map their own oral storytelling traditions about places of significant meaning or value to them.

bperlik:
I can do this.

mirandawang:
@bperlik Here's the introduction text :) 
```Terrastories are audiovisual recordings of place-based storytelling. This offline-compatible application enables local communities to locate and map their own oral storytelling traditions about places of significant meaning or value to them.```
Thanks for claiming this issue!

mirandawang:
@bperlik Sorry, I misread your message! The introduction text should go into the card partial, under "What's a terrastory?". Currently the placeholder is Lorem Ipsum. Thanks again!

nickmjones:
Candidate to move into "MVP for Matawai" milestone?

nickmjones:
Marking this closed. These changes should be in `master` now.

mirandawang:
Reopening this issue since it seems to still be addressed in https://github.com/rubyforgood/terrastories/pull/79

nickmjones:
Closed in #89

PR #79, opened 2018-08-16
Closes issue #76

nickmjones:
Replaced the intro text in the map legend/overlay.

### DIFF
--- a/rails/db/schema.rb
+++ b/rails/db/schema.rb
@@ -33,16 +33,6 @@ ActiveRecord::Schema.define(version: 2018_06_10_154604) do
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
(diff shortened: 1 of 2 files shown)


rails/db/schema.rb
```diff
@@ -33,16 +33,6 @@
...
-  create_table "media", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
-    t.bigint "story_id"
-    t.string "media_type"
-    t.string "description"
-    t.string "name"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
-    t.index ["story_id"], name: "index_media_on_story_id"
-  end
-
```
mirandawang: We probably need to keep this in the schema right? We want to be able to associate media like videos with stories.

PR #84, opened 2018-08-18
[83] Improve look and feel of map balloons

nickmjones:
- Updates to the way SCSS was organized
- Minor reskin of the mapbox popups. I feel like there's probably a better way to do that than to overwrite the CSS, but it works and doesn't require `!important` to work, so it should be ok for production.


nickmjones:
I've restored the part of schema.rb that went missing inexplicably, so this should be good to go. I'll make sure to follow the contribution guidelines more closely going forward also!

rails/db/schema.rb
```diff
@@ -114,7 +114,6 @@
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
   end
 
-  add_foreign_key "media", "stories"
```
mirandawang: Could you add this line back too please? Thanks!

mirandawang:
One more line missing from the schema.rb file, but otherwise everything else is looking really good! :)

nickmjones:
Ok, this should finally be ready to go. Could someone take a look and see how all that stuff ended up getting excised from `schema.db`? My git foo is not the strongest, and it seems like I accidentally picked up an upstream change of some kind since it wasn't just random deletions.

Bootjack:
I'm not sure what would have modified `db/schema.rb` other than manually changing it. That could have happened while resolving a merge conflict, but I don't see what would have triggered a conflict on that file. I'll be running through a few clean install attempts trying to get this running on the POSM hardware, so I'll keep an eye out for any setup steps that could cause this.

It looks like you merged master into your feature branch, which is kind of backwards from what you probably wanted to do. In general, you just want to have one merge commit when your feature is finished and you merge that feature branch into master. To keep your feature branch up to date with master you should rebase your feature branch on to master with `git rebase master feature-branch-name`. That will go through the same conflict resolution steps as a merge, but you don't end up with extra merge commits each time you do it.

To clean up that extra merge at this point, you probably want to do the following:
1. Make sure your local master is in sync with github
  a. `git checkout master`
  b. `git reset --hard origin/master`
  c. `git pull origin master`
1. Start an interactive rebase `git rebase -i master nmj-gh83`
2. Inside the interactive rebase, remove the line `pick 787fb116 Merge branch 'master' ...`
3. Save, finish the rebase

nickmjones:
@Bootjack thank you, sir. Really appreciate you taking the time on that. I'll follow those steps and try and turn this around quickly. I want to start a few other tasks today, so I'm hoping to clear the decks.

nickmjones:
Closing this and deleting the branch. I'll reupp the PR today.

Issue #80, opened 2018-08-17
Responsive design for the app!


mirandawang:
Put basic responsive styles into place, including a minimal navigation that will work properly on mobile devices.

Issue #67, opened 2018-06-15
Make the search icon clickable and turn it into a search bar


mirandawang:
When you click on the search icon, it should turn the bar into a search bar where the user can type in text. It hides the filter dropdowns.

![search example](https://user-images.githubusercontent.com/8590508/41489330-9c280a82-70bd-11e8-9b7e-5b433b63407e.png)

mirandawang:
@mdworken Sure! We would love if you could contribute! Please view our CONTRIBUTING.md guidelines for more details on the steps :)

Issue #62, opened 2018-06-15
Make the introduction disappear on scroll, keep the search bar & logo sticky


mirandawang:
Update our CSS so that when the user scrolls in the card, the introduction disappears (but reappears when you scroll back up) and the search bar & logo are in a fixed position. 

To achieve this you could make only the container under the logo scrollable, and set the position of the search/filter bar to be fixed. 

![list view](https://user-images.githubusercontent.com/8590508/41486125-3d77293e-70b1-11e8-8bb9-3f683489b6de.png)

PR #93, opened 2018-08-30
UI related stories

nickmjones:
This work (hopefully) closes obviates several stories related to the map card/sidebar.
- Deals with search field (#67)
- Handles the behavior of the intro and several other pieces of content when scrolling (#62)
- Lays the groundwork for mobile/responsive by giving us a tray for the sidebar instead of the floating palette (#80)

I realize this is a big PR that touches a lot of things, and I don't necessarily require approval now since this is my "ui branch". Just wanted everyone to be aware so I can gather reactions, and keep pushing other items into this branch.

screenshot_2018-08-30 app 2

screenshot_2018-08-30 app

screenshot_2018-08-30 app 1


rails/app/javascript/components/IntroPopup.jsx
```diff
@@ -0,0 +1,34 @@
+import React, { PureComponent } from "react";
+import PropTypes from "prop-types";
+
+class IntroPopup extends PureComponent {
+
+  constructor(props){
+    super(props);
+    this.state = { isPopped: true }
+    this.handleIntroPopup = this.handleIntroPopup.bind(this);
```
mirandawang: How do you feel about using the fat arrow function syntax to get around having to bind every function to `this`?  I'm not sure if there's really a difference between the two ways, but it could save you some typing in the future :) 

So in the function declaration, instead of 
```
handleIntroPopup () {
  // stuff
}
```
you can do
```
const handleIntroPopup = () => {
  // stuff
}
```
and you'll have access to `this` context, so you won't need to bind it in the constructor.

nickmjones: Absolutely, @mirandawang! Good catch and I learned something :)

mirandawang: Thanks for updating! :) So you should be able to delete all the `.bind()` references now!

nickmjones: Done and done :)

Issue #88, opened 2018-08-21
Add a permission level field to Stories


mirandawang:
Some stories can be viewed by any user, anonymous or not. Other stories can only be viewed if the user is logged in. Let's add a field to Stories that allows Editors to set the permission level (public or private).

Issue #86, opened 2018-08-21
Create pundit policy for editors


mirandawang:
Add a policy in Pundit for Editors
Editors can be defined as a user who can see all stories, add new stories, and edit stories
Make sure to include the check for this policy when viewing, adding, and editing stories

bperlik:
I would like to work on this.

PR #101, opened 2018-09-23
[88][86] Add permission level to stories, role to users, add policy for viewing Stories, and clean up administrate dashboards

mirandawang:
This PR adds two new fields as enums:
permission_level on Story (:anonymous, :user_only, :editor_only)
role on User (:user, :editor) 

I also included a policy for Stories and updated the helper method for viewing only the Stories that the current user has permission for.

This PR also updates and cleans up the Administrate dashboards so they work without any errors! This should make the dashboards more readable and easier to navigate and create new stories. You can create users, delete users, and edit their passwords. (Unsure why the password field doesn't actually show up in the dashboard - it could be a security thing so I didn't look into trying to fix that yet.) You can also set the permission level on a Story which will determine which can be viewed in the map page. 

There is also some new authentication logic for viewing the admin page. You will need to be logged in, and have the Editor role to view the admin dashboards. I included an Editor user in the seeds file for development purposes. 

I also temporarily changed the Welcome page to display the current user's email address if logged in, a link to edit their own profile or log out, and if not logged in, a link to sign up or log in. This is a temporary change to help with development, I plan on creating a new issue to clean this up in the UI. 

Things I would like to add next:
Update the card component to display the actual user's information, with log out and edit profile links
Update the welcome card to display the admin information (including a link to the admin dashboard) if the user is logged in as an admin/editor.

### DIFF
--- a/rails/app/controllers/admin/application_controller.rb
+++ b/rails/app/controllers/admin/application_controller.rb
@@ -6,10 +6,11 @@
 # you're free to overwrite the RESTful controller actions.
 module Admin
   class ApplicationController < Administrate::ApplicationController
+    before_action :authenticate_user!
     before_action :authenticate_admin
 
     def authenticate_admin
-      # TODO Add authentication logic here.
+      redirect_to '/', alert: 'Not authorized.' unless current_user && current_user.editor?
     end
 
     # Override this value to specify the number of elements to display at a time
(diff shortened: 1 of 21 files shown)


mirandawang:
Closes Issue #88 and #86

PR #106, opened 2018-09-25
asset pipeline fixes to make it happy in Heroku

robbkidd:
* set JS compressor to use harmony mode for ES6

Resolves the following error when compiling assets:

    Uglifier::Error: Unexpected token: punc ()). To use ES6 syntax,
    harmony mode must be enabled with Uglifier.new(:harmony => true).

* fix import to match case with filename, `IntroPopup != introPopup`
* updated `image_tag` calls to use the complete image filename to avoid asset pipeline lookup misses in production

### DIFF
--- a/rails/app/assets/stylesheets/components/_components.scss
+++ b/rails/app/assets/stylesheets/components/_components.scss
@@ -8,4 +8,4 @@
 @import 'speaker';
 @import 'welcome';
 @import 'balloon';
-@import 'IntroPopup';
+@import 'introPopup';
--- a/rails/app/views/home/_card.html.erb
+++ b/rails/app/views/home/_card.html.erb
@@ -1,6 +1,6 @@
 <div class="card">
   <div class="logo">
-    <%= image_tag 'logocombo', alt: 'Terrastories' %>
+    <%= image_tag 'logocombo.svg', alt: 'Terrastories' %>
   </div>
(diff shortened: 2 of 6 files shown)


rails/app/assets/stylesheets/components/_components.scss
```diff
@@ -8,4 +8,4 @@
 @import 'speaker';
 @import 'welcome';
 @import 'balloon';
-@import 'IntroPopup';
\ No newline at end of file
+@import 'introPopup';
```
robbkidd: Essentially, this import and the filename for the component should match. `IntroPopup`, `introPopup`, `intro_popup`, whatevs. It'd be OK to go with this change to the import, or rename the file, or both. Just gotta pick one.

rails/app/views/home/_card.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="card">
   <div class="logo">
-    <%= image_tag 'logocombo', alt: 'Terrastories' %>
+    <%= image_tag 'logocombo.svg', alt: 'Terrastories' %>
```
robbkidd: Filename needs to match, otherwise it causes a missed lookup in the asset pipeline when running in production.

robbkidd:
Nope. This set of changes updates code to be in line with existing filenames.

PR #107, opened 2018-09-25
Km/reactify all the things

kalimar:
This PR converts all views to React components. This is intended to make it easier to communicate stage changes between components of the app.
* Created `App` and `Map` components.
* `App` is a parent component which makes communication between `Map`, `Popup` and `Card` easier.

This breaks the jQuery code that exists. For example https://github.com/rubyforgood/terrastories/blob/km/reactify-all-the-things/rails/app/javascript/components/Map.jsx#L55

Added a method to pass story element https://github.com/rubyforgood/terrastories/blob/master/rails/app/javascript/components/StoryList.jsx#L16 geographic coordinates to the `Map` component.


kalimar:
Oops - squashing these commits. That [Do NOT MeRGE] is pretty ugly!

kalimar:
Ok, squashed and ready for review 😸

PR #108, opened 2018-09-25
Rudo design

rudokemper:
This one contains a few different kind of commits:

Stylistic
-Redesign of the welcome page (and added some Matawai text)
-Adding title of application and favicon (closes issue 96)
-Stylistic changes to card, map, intro popup view
-Updated some of the graphics
-Hiding search bar for now

Small fixes
-Fix to CSS where introPopup.scss was improperly capitalized
-Fix to docker-compose.yml to enable Jason's latest PR to work on Windows

Map content
-Updated shapefiles, and new style.json

The one issue I'm still having is the one related to rendering Sprites for points in the map, like I mentioned (couldn't figure out how to define it properly in local style.json). Now somebody can take a look at that. The two SVG sprites are in tileserver/data/sprites. You can see how they're supposed to be rendering in http://www.amazonteam.org/maps/matawai-oral-histories/map.html

capture2
capture1


docker-compose.yml
```diff
@@ -7,7 +7,7 @@ services:
     networks:
       - railroad
     volumes:
-      - ./data:/var/lib/mysql
+      - dbdata:/var/lib/mysql
```
robbkidd: 👍 Stow the database in a docker volume will be more better.

PR #110, opened 2018-09-26
Add convenience scripts to ease development and deployment

Bootjack:
This is an almost-faithful implementation of [these github conventions](https://github.com/github/scripts-to-rule-them-all). It adds several utility scripts that normalize some of the more arcane docker commands into more readily understood ones.

`bin/bootstrap`
Installs dependencies. For this project, that's just docker. (Also, this script currently only works in a Debian-flavored Linux environment. Sorry!)

`bin/setup`
Performs first-time setup. For this project, that means building all images and running tile-builder for the first time.

`bin/update`
Performs appropriate tasks after a fresh pull from github. For this project, that means rebuilding docker images so that gems and node modules are cached.

`bin/server`
Run the server and log output from nginx. Once started, the app should be available at [localhost:3000](http://localhost:3000).

`bin/logs`Show the live logging output from the running rails container. This will fail if the container isn't running.

`bin/console`: Start an interactive shell session on the running rails container. This will fail if the container isn't running.

Additional scripts to be aware of:
`bin/start` and `bin/stop`: Start and stop the app much like `bin/server`, but do so in detached mode. This is my personally preferred way to run the app. Most people would expect some ongoing output after starting their sever, so I'm keeping these distinct from `bin/server`.

`bin/install` and `bin/uninstall`: Intended for use in deployed environments, not development. These expect to have root permissions on an Ubuntu system. The install script also copies the upstart/initctl configuration to run terrastories as a service whenever the docker service is started. These are untested and potentially dangerous scripts. Please don't run them on your development machine.

Despite my stern warning about the install script, I do need some help testing it. It is intended to provide a one-line self-installation option to make it easier to install the application on a NUC or similar hardware. ~~The command for testing that installation is:~~
~~curl https://raw.githubusercontent.com/rubyforgood/terrastories/convenience-scripts/bin/install | sh~~
[_nvm don't do this; see comment below_]

I haven't been able to test this despite having the POSM image up and running locally with VirtualBox simply because VirtualBox isn't letting the VM use my network connection. Any help is appreciated!

### DIFF
--- a/conf/terrastories.conf
+++ b/conf/terrastories.conf
@@ -1,0 +1,7 @@
+description "Start the Terrastories server"
+author "Jason Hinebaugh"
+
+start on started docker
+stop on stopping docker
+
+exec /bin/terrastories-start
--- a/script/console
+++ b/script/console
@@ -1,0 +1,2 @@
+#!/bin/bash
+docker-compose exec rails bash
--- a/script/logs
+++ b/script/logs
@@ -1,0 +1,2 @@
+#!/bin/sh
+docker-compose logs -f rails
--- a/script/prod/bootstrap
+++ b/script/prod/bootstrap
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Install docker if not already installed
+if [ "$(which docker)" == "" ]; then
+  sudo apt-get install -y apt-transport-https ca-certificates software-properties-common
+
+  # First, add the GPG key for the official Docker repository to the system:
+  curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+
+  # Add the Docker repository to APT sources:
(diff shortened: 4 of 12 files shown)


robbkidd:
What do you think about separating the scripts that are good for development environment from the scripts that are intended for the sparse-internet hardware? I get the sense that the [github conventions](https://github.com/github/scripts-to-rule-them-all) are very much intended for development and CI environments.

Bootjack:
@robbkidd I like that. I think the convention uses a `scripts` dir and I picked up the `bin` name from a slightly different convention at my office. Thoughts on which dir names could help indicate the distinction? Maybe `deploy` or `prod` for the non-dev scripts? Or maybe `scripts/dev` and `scripts/prod`?

PR #123, opened 2018-09-30
Make containers restart when computer reboots

kalimar:
I've added `restart: always` to the containers that need to be on for the app to work. This was my 2 cents solution so definitely open to feedback on a better way of doing this.


kalimar:
Did I do this right? @seanmarcia @robbkidd ?

PR #129, opened 2018-10-01
Mapbox switcher

lauramosher:
## What

This codifies the logic for switching between the local tile style and the Mapbox style.

## Notes

This sets up and relies on ENV to be set. Since this is deployed to Heroku, we'll need to configure Heroku for two new environment variables:

- MAPBOX_ACCESS_TOKEN
- MAPBOX_STYLE

See [Slack Convo about Mapbox Style](https://rubyforgood.slack.com/archives/CAH9XQX2A/p1538403515000100?thread_ts=1538403332.000100&cid=CAH9XQX2A) and [Slack Convo about Access Token](https://rubyforgood.slack.com/archives/CAH9XQX2A/p1538403779000100?thread_ts=1538403332.000100&cid=CAH9XQX2A) for the correct string values.

### DIFF
--- a/rails/app/views/home/_home.json.jbuilder
+++ b/rails/app/views/home/_home.json.jbuilder
@@ -12,3 +12,5 @@ json.stories stories do |story|
 end
 json.logo_path image_path("logocombo.svg")
 json.user current_user
+json.mapbox_access_token ENV["MAPBOX_ACCESS_TOKEN"]
+json.mapbox_style ENV["MAPBOX_STYLE"] || "/tiles/styles/basic/style.json"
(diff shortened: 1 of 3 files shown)


rails/app/views/home/_home.json.jbuilder
```diff
@@ -11,4 +11,6 @@ json.stories stories do |story|
   end
 end
 json.logo_path image_path("logocombo.svg")
-json.user current_user
\ No newline at end of file
+json.user current_user
+json.mapbox_access_token ENV["MAPBOX_ACCESS_TOKEN"]
```
robbkidd: Is `nil` a sensible default for this token?

lauramosher: 🤔 other than perhaps an empty string, but Mapbox doesn't really care about this value if you're not calling Mapbox. Null (in JS) will be sufficient for local development.

PR #134, opened 2018-10-02
Add internationalization

### LINKED ISSUE #29: Internationalization and Localization
The native language of the people of Suriname is Matawai, which none of us speak. Well, Rudo does.

Add support for localization so that we can build the app in English and someone (Rudo) can translate phrases in our app to Matawai.

Add support for switching locales, probably with support for routes beginning with `/en/` and `/srm/` to distinguish between the two. (It seems "srm" is the official language code for Saramaccan, of which Matawai is a dialect, I guess.)

http://guides.rubyonrails.org/i18n.html

lauramosher:
## What

This adds internationalization to the repository.

Since we're using Rails + React, this adds a gem called `i18n-js` that allows the usage of Rails i18n in the JavaScript files.

I've scoped the locale so if it's provided in the URL, it automatically switches language locales. Currently, by default, it's set to use Matawai / Suriname.

This can be achieved two ways:

- `terrastories.io/en`
- `terrastories.io/?locale=en`

They both will behave the same*. :tada:

## *Notes

1. Admin pages do not work in the Matawi language (with the exception of "Back to Map" link).
2. Admin pages are *not* scoped, so they are required to use the `?locale=en` version.

## Demo

| en | srm |
| --- | --- |
| image | image |
| image | image |
| image | image |

Closes #29


lauramosher:
Added a basic language switcher

image

PR #137, opened 2018-10-03
Small design fixes

rudokemper:
* updated map marker

* adding logos for ACT and our Matawai partner organization on the front page

* small CSS tweaks here and there

* few minor updates to I18n

* cropping speaker photos in Card view and new thumbnail (mostly covers https://github.com/rubyforgood/terrastories/issues/102)

* saved .sh scripts in UNIX format (for Windows operation)


rails/app/assets/stylesheets/components/card.scss
```diff
@@ -146,11 +146,19 @@ $el-shadow: 0 1px 4px #bbb;
...
         p {
-          font-size: 0.7rem;
+          font-size: 0.85rem;
           line-height: 1.25;
           margin-top: 1rem;
           @include ui-font;
         }
+		.speakers {
+			img { 
+				width: 100px;
```
lauramosher: This is probably fine as a stop gap, but if someone uploads a non square image, this will stretch / squash (matter of perspective) rectangular images.

image

rails/app/views/welcome/index.html.erb
```diff
@@ -27,4 +27,7 @@
       <%= link_to "en", root_path(locale: "en") %>
     </div>
   </div>
+  <div class="logo">
+	<%= image_tag 'act-matawai.png', alt: 'Amazon Conservation Team and Stichting voor Dorpsontwikkeling Matawai' %>
+  </div>
```
lauramosher: This chunk may need additional styles, or be hidden, when the screen is tiny. It does some really strange things to the primary welcome bubble.

rails/app/assets/stylesheets/components/welcome.scss
```diff
@@ -8,6 +8,18 @@
   align-items: center;
   text-align: center;
   
+  .logo {
+  
+		 justify-content: center;
+		align-items: center;
+		text-align: center;
```
lauramosher: The above 3 styles are inherited from `.welcome` in the markup; you can rely on those instead.

rails/app/assets/stylesheets/components/welcome.scss
```diff
@@ -8,6 +8,18 @@
...
+	
+		img {
+			height: 50%;
```
lauramosher: Probably don't need a height here; setting the width to the desired size will ensure the logo remains the proper ratio.

rails/app/assets/stylesheets/components/card.scss
rudokemper: Very true. We've got all square images, so it will work for the immediate future (i.e. tomorrow), but maybe we can build this out more.

rails/app/views/welcome/index.html.erb
rudokemper: Thanks for catching this. I worked on the .scss file some more.

rails/app/assets/stylesheets/components/welcome.scss
rudokemper: Fixed in updated style sheet. Thanks!

rails/app/assets/stylesheets/components/welcome.scss
rudokemper: Perfect. Fixed.

PR #144, opened 2018-10-04
Try virtualized list

mirandawang:
This PR uses React Virtualized List to lazy load the stories so the DOM is not overwhelmed with media attachments. I was having issues figuring out the pagination so I used this library instead - however, this keeps intact the filtering and permissions logic.


kalimar:
@Bootjack is gonna love this!

PR #179, opened 2018-11-09
update seeds to create a _valid_ admin user

robbkidd:
password_confirmation also needs to be set for the validations to be happy. Otherwise a super unhelpful 'password.too_short' validation error occurs.

### DIFF
--- a/rails/db/seeds.rb
+++ b/rails/db/seeds.rb
@@ -45,8 +45,8 @@ Story.find_or_create_by(title: "Fa di Kwinti nengeb bi feti ku Matawai sembe",
                      point: flaming_forest,
                      permission_level: 1)
 
-User.where(email: 'admin@terrastories.com')
-    .first_or_create do |admin|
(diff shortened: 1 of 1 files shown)


rails/db/seeds.rb
```diff
@@ -45,8 +45,8 @@
                      point: flaming_forest,
                      permission_level: 1)
 
-User.where(email: 'admin@terrastories.com')
-    .first_or_create do |admin|
-      admin.password = 'password',
-      admin.role = 1
-    end
+User.where(email: 'admin@terrastories.com').first_or_create do |admin|
```
robbkidd: Also to be considered is updating `first_or_create` to the bang-method version `first_or_create!`. This would have the `db:seed` rake task error with an exception if an `admin@terrastories.com` user cannot be found or _successfully created_ (i.e. this admin user seed falls out-of-sync with user model validations).

PR #184, opened 2018-12-01
WIP 130 Logout

LizPrescott:
This is an attempt at fixing the logout button from the maps page. Read up a lot on the issue of sending a delete request to the devise sign_out end point, and it is not straightforward. The code in this pr make an attempt with axios, but the button still appears to not do anything. I can see in the console that it is returning a 404 error, even though the headers seem to be the same as those in the main log out button on the welcome page. Hoping that someone with more experience in REACT or devise will be able to see right away what I'm missing here. 
I did look at this [article](https://blog.andrewray.me/how-to-set-up-devise-ajax-authentication-with-rails-4-0/), but I think it's incorrect. Looking at the Devise source code, it seems that the end point already accepts json, so that should not be the issue. Other potential resources:
https://makandracards.com/makandra/53562-devise-invalidating-all-sessions-for-a-user
Devise seems to be doing a lot under the hood that I haven't quite accounted for in my axios code. halp!

### DIFF
--- a/rails/app/controllers/application_controller.rb
+++ b/rails/app/controllers/application_controller.rb
@@ -19,4 +19,8 @@ class ApplicationController < ActionController::Base
     flash[:alert] = "You are not authorized to perform this action."
     redirect_to(root_path)
   end
+
+  def after_sign_out_path_for(resource_or_scope)
+    root_path
+  end
 end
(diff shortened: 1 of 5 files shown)


mirandawang:
Closing for now, we'll reprioritize this in the future! Thank you for your work! <3

PR #185, opened 2018-12-01
Update documentation to use convenience scripts

Bootjack:
(Sorry, it looks like my Atom linter plugin had some opinions about these files in addition to my actual changes.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,117 +1,144 @@
 ![Terrastories](https://www.amazonteam.org/wp-content/uploads/2018/09/logo-1170x164.png)
 
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,117 +1,144 @@
...
 ```
-$ docker-compose build
-  ...
-> mariadb uses an image; skipping
-  ...
-> Successfully built 0123456789
-> Successfully tagged terrastories/tilebuilder:latest
-> tileserver uses an image; skipping
-> rails uses an image; skipping
+$ script/setup
```
csexton: For this and the other script examples ya might want to include the `./` at the beginning of this example for folks that don't have `.` in their `PATH`

E.g.:

```
$ ./script/setup
```

PR #186, opened 2018-12-05
[176] Allow for more granular permissions based on user access control groups

Bootjack:
This adds a new concept in the `Demographic` model which links users and stories through an access control grouping. I'm open to suggestions for the name, but it seems like these will mostly be based on real-world social groupings, such as elders from a certain village or descendants of one individual. The term "demographics" felt like a decent fit. This is a different concept from both user roles and story permissions. The three complementary concepts we'd have after this PR are:

Story permissions - an individual story is marked as either:
  - `:anonymous` meaning public, or
  - `:user_only` meaning any signed-in user may see it

User roles - an individual user has a single role that's one of:
  - `user` basic signed-in user, may view some restricted content
  - `editor` same as user, and also may edit story and location content
  - `admin` same as editor, and also may edit other users ~~(just added as an option, not yet implemented/enforced)~~

Demographics - a group of users and stories that defines restricted access levels using the following data (basically describing two new HABTM relations)
  - `name` a human-readable plain-language name for the group
  - `users` a list of users included in this group
  - `stories` a list of stories this group has access to

---

Punch List

  - [x] Create the Demographic model
  - [x] Add relationships between Demographic, User, and Story models
  - [x] Customize the demographic dashboard in administrate
  - [x] Filter stories based on Demographics
  - [x] Remove the direct User-Story access control association
  - [x] Replace the `Story.permission_level` field with `Story.is_public`
  - [x] Add pundit policies so that editors:
    - [x] may not modify User models
    - [x] may not view Demographics they don't belong to
    - [x] may not add stories to Demographics they don't belong to
    - [x] can only see appropriate Users and Stories when modifying Demographics
    - [x] may not add Demographics outside of their permissions to Stories they have access to
  - [ ] Automatically add the user who creates a Demographic to that Demographic (otherwise only
admins can access it)
      - [x] add the current user to the demographic
      - [ ] prevent duplicate records
  - [x] Do not list Demographics under Stories that when an editor doesn't belong to them
  - [ ] Prevent limited Demographics scope of editor from deleting existing Demographics when updating Stories

---

For Discussion

I think we can streamline these concepts a bit, but we probably do want some form of all three. Check out this branch, go to the new Demographics dashboard in the admin and see if creating demographic groups and associating users and stories seems logical. Here are some specific questions I have for the team.

1. Story permissions is currently a text field at `Story.permission_level` but I think at this point it only needs to be a boolean indicating whether a story should be shown to the public, something like `Story.is_public`.
2. Similarly, should we expect that the combination of user role plus demographics will always result in the correct editing permission for stories. In other words, if a user is an editor, should they have editing access to all stories their demographics allow them to see? Or should there be a distinction between view-only demographics and editor-specific demographics?
3. Does the `admin` user role make sense? My expectation would be that members of the ACT who are setting things up (plus maybe a few tech-savvy locals) would be the only admins. Everyone else would go to that group to get elevated editing permissions or to modify demographics for access control.
4. I also modified the models to allow for direct user-to-story access levels, but I'm curious if this is really needed or if it just adds extra complexity and confusion. The same could be accomplished by creating a demographic group with only one user. If that's a rarity it should be fine; if it's common the number of single-user demographics could become a problem. Note, there's no dashboard UI for setting up these direct user-story relationships yet.


mirandawang:
AWESOME!!! :100: :tada: 

Thanks for getting this started!

Discussion points:
1) I think updating that filed to a boolean `is_public?` makes total sense here, since we only really have that distinction now for Stories.
2) Editing and viewing to me are distinct. Viewing stories is on a user level, but some users also have the ability to add/update/delete stories. That also means they have access to the /admin backend. There could be an "editor-only" demographic for stories which aren't ready to be published yet, or something.
3) admin makes sense to me! In the future these would be the users that could add and update custom fields for the models too. 
4) Maybe @rudokemper can help us answer that question, is it more common that stories would be visible to certain groups or usually just individuals?

Bootjack:
Thanks for the feedback, @mirandawang! To clarify my second question, I really just mean should an editor be able to edit every story they can view? I think we're agreeing that they should, but I wanted to make sure.

rudokemper:
Brilliant stuff Jason!! 👍👍👍

1. I would agree here.
2. Yeah, I mean, theoretically there could be a scenario where a community member gets to view stories X, Y, and Z, but only have permission to edit X. Let's say that they are a hunter and story A is a hunting story, whereas Y and Z are stories told by elders which they should be able to hear, but not modify. But I really can't imagine this playing out in reality, certainly not anywhere where we work for instance. Let's just keep it simple.
3. Fully agreed here and also with @mirandawang's suggestion about adding and custom fields for the models in the future.
4. My feeling is that the demographic groups (with one-user groups if necessary) are perfectly sufficient, and that having the additional direct-to-user setup could likely result in more confusion.

mirandawang:
Cool so, to sum up: 
* Editors can edit or delete any story that they can view / they have created.
* Story permission levels will be based on Demographic groups. For the rare case of individual level permissions, we can model this as a one person group. 

@rudokemper @Bootjack @kalimar does that all sound right?

Bootjack:
@kalimar In my defense, jQuery was already a part of administrate 😜

Bootjack:
Also, some general explanation: Those last few commits (including the jQuery stuff) are in support of honoring pundit policies in the multiple-select widget for Stories/Demographics/Users in the administrate forms. To do that I created a custom field `HasManyScopedField` that subclasses the normal `Field::HasMany` from administrate. That felt a little hacky because I had to copypasta two methods from the base class and tweak them.

Even more hacky is that in order to pick up the styles and js from the base field I had to run their rails generator that copied all the assets for all fields and dumped them into our app. I'm sure a more clever dev could have come up with a cleaner way to do that. Eventually, though, I think the in-map editing will supplant this entire administrate interface so I'm not super worried.

Bootjack:
lol, nvm. @kalimar 's nerd snipe was successful

kalimar:
You know me too well @bootjack. You know me too well

mirandawang:
:eyes:

mirandawang:

PR #188, opened 2019-01-05
Display the speaker name in each story

AliTopal89:


rails/app/javascript/components/StoryList.jsx
```diff
@@ -63,7 +63,8 @@ class StoryList extends Component {
             style={style}
           >
             <div className="speakers">
-              <img src={story.speaker.picture_url} alt={story.speaker.name} title={story.speaker.name}/>
+              <img src={story.speaker.picture_url}/>
```
mirandawang: We might want to keep the alt and title properties here for accessibility reasons. What do you think?

AliTopal89: I checked it out with it and without it. I really didn't see the difference but I have no problem with adding that back in.

mirandawang: I think when you use screen readers, that’s how it knows how to describe the images. I haven’t tested it out myself though.

AliTopal89: oh ok I understand what you are saying. Adding that back in now.

mirandawang:
Thanks Ali looks great!!

PR #190, opened 2019-01-24
fixes admin user creation in the seeds

robbkidd:
THE COMMAS! The commas turned the three statements into a single statement that returned an array:

    ["password", "password", 1]

So nothing got set on the attributes for the admin object and so the user was invalid because the password was too short.

This change removes the commas and updates the code to use `find_or_create_by!` as [documented in the Rails guide](https://guides.rubyonrails.org/v5.2/active_record_querying.html#find-or-build-a-new-object). The band-method version was chosen to that seeding the database would fail if it can not find the admin user or create a new valid one. Eventually this admin seed will need to change to not expect all installations to use the same email for the admin user.

### DIFF
--- a/rails/db/seeds.rb
+++ b/rails/db/seeds.rb
@@ -45,8 +45,8 @@ Story.find_or_create_by(title: "Fa di Kwinti nengeb bi feti ku Matawai sembe",
                      point: flaming_forest,
                      permission_level: 1)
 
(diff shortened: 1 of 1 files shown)


robbkidd:
Nevermind! I see this change is [already in master](https://github.com/rubyforgood/terrastories/commit/3214bc5bbba56e48b4329903163142f696198469).

PR #207, opened 2019-03-08
Add migration to fix coordinates

rudokemper:
Addressing this issue: https://github.com/rubyforgood/terrastories/issues/204

Added a migration to set the precision/scales to 10/6 for both `lat` and `lng` columns.

Thanks @robbkidd for thinking through this one! 👍


robbkidd:
I'm going to think through it a little bit more. I'm tinkering today on some sample data to try to find if there are any edge cases in the precision/scale change.

PR #208, opened 2019-03-09
set nginx container to not start until rails is listening

robbkidd:
Change the nginx service in the docker-compose configuration to build the nginx image instead of using the one directly from Docker Hub. Added [dockerize](https://github.com/jwilder/dockerize) to our nginx Dockerfile so that it will handle rendering the template of the nginx configuration *and* wait to start the nginx service until something is listening on the rails host at tcp:3000.

Also, the nginx service does not need to know the `NODE_ENV` and `RAILS_ENV` values, so they were removed.

Updated the logs script to include logs from nginx so a person might see that nginx is waiting for rails to be available.

Update logs and start scripts to use the long form of command switch options so that they are less arcane to readers.

With this change, `http://localhost:3000` will look like this until Rails starts up:

> ### This page isn’t working
>
> **localhost** didn’t send any data.
>
> ERR_EMPTY_RESPONSE

On the command prompt, things would look like:

<details><summary> > script/start </summary>

```
terrastories_mariadb_1 is up-to-date
terrastories_tileserver_1 is up-to-date
Creating terrastories_rails_1 ... done
Creating terrastories_nginx_1 ... done
```

</details>
<details><summary> > script/logs </summary>

```
Attaching to terrastories_nginx_1, terrastories_rails_1
rails_1        | Removing existing yarn modules from host
nginx_1        | 2019/03/09 21:24:59 Waiting for: tcp://rails:3000
nginx_1        | 2019/03/09 21:24:59 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:00 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:01 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:02 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:03 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:04 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
nginx_1        | 2019/03/09 21:25:05 Problem with dial: dial tcp 172.23.0.3:3000: connect: connection refused. Sleeping 1s
[...]

PR #210, opened 2019-03-11
display an better error message when nginx 502s

robbkidd:
502 most likely means the Rails process is not responding (down or still starting up). Display a meaningful message instead of the enigmatic "Bad Gateway" no soup for you message.

Changed the nginx service to be built from an image so that the error page could be added directly to the image. Also added dockerize to handle the template rendering for the nginx config.

Removed NODE_ENV and RAILS_ENV environment variables as unnecessary for the nginx service.

This is an alternative to #208 (the one [mentioned there as Better Error Message](https://github.com/rubyforgood/terrastories/pull/208#issuecomment-471228562)).

Looks like this currently, but obviously the words are negotiable:

Screen Shot 2019-03-11 at 4 36 16 PM


robbkidd:
Something to consider: as written at the moment, this page will appear as-is for field units when the Rails process causes a 502 in nginx.

mirandawang:
What do you mean by field units?

robbkidd:
The Terrastories instances running standalone "in production" in communities. It's my understanding that those installs use this docker-compose config and these dockerfiles (or at least the images generated by them).

With this change, a Terrastories NUC in a community would display this error page until the Rails web service was up and responsive.

Bootjack:
Hey, I had started something very similar here https://github.com/rubyforgood/terrastories/compare/69-startup-message although it looks like you might have gotten to a more complete point. I did try to add polling to the "sorry" screen so that it would automatically redirect to the login page as soon as it get's a successful response from the rails app. I'd love to hear your thoughts on that.

mirandawang:
https://github.com/Terrastories/terrastories/pull/208

mirandawang:

PR #212, opened 2019-03-30
Add guard, get specs working

baweaver:
This change adds guard for automatically running tests, and temporarily
disables one failing integration test until we decide what we want to
replace it with.


rails/spec/controllers/speakers_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
baweaver: The previous file was misnamed

baweaver:
Whoops, resubmitted to the rubyforgood fork.

PR #213, opened 2019-03-30
#5 Migrations update to stories

s-nicas:
Add date interviewed, interviewer and language to stories table
Add has many interview locations to stories model


baweaver:
You might want to target this off of the `rubyforgood/terrastories` repo

PR #221, opened 2019-04-24
Allow sidebar to be zoomed over minimized sidebar

### LINKED ISSUE #203: [UI] When sidebar is minimized, can't zoom in or out when moused over that area
In the Terrastories UI, it is possible to minimize the sidebar using the arrow bottom at the top right of the sidebar. 

Once the sidebar has been minimized, and you try to zoom in or out on the map using the middle mouse button in the exact area where the sidebar used to be, nothing happens.

alindeman:
Fixes #203 

Previously, it was not possible to zoom in/out over the area where the collapsed sidebar was. This mitigates the issue by reducing the width of the collapsed sidebar just to the area necessary to display the tab.

Screenshot 2019-04-23 at 21 51 22

There's still `2rem` of area that's not zoomable, but I couldn't figure out a way to get around that without much more invasive changes. Feedback welcomed though.

### DIFF
--- a/rails/app/assets/stylesheets/components/card.scss
+++ b/rails/app/assets/stylesheets/components/card.scss
@@ -20,8 +20,10 @@ $el-shadow: 0 1px 4px #bbb;
   }
 
   &.offCanvas {
+    max-width: 2rem;
+
     .card {
-      left: -100%;
+      display: none;
     }
     .tab {
       &:after {
--- a/rails/app/assets/stylesheets/components/card.scss
+++ b/rails/app/assets/stylesheets/components/card.scss
@@ -20,10 +20,10 @@ $el-shadow: 0 1px 4px #bbb;
   }
 
   &.offCanvas {
-    max-width: 2rem;
+    pointer-events: none;
 
     .card {
-      display: none;
+      left: -100%;
     }
     .tab {
       &:after {


rails/app/assets/stylesheets/components/card.scss
```diff
@@ -20,8 +20,10 @@ $el-shadow: 0 1px 4px #bbb;
   }
 
   &.offCanvas {
+    max-width: 2rem;
+
     .card {
-      left: -100%;
+      display: none;
```
mirandawang: Would you mind updating this to use a wrapper element that has 0 width/height and `overflow:hidden`, then set the card element to `visibility:hidden` instead of `display:none`? This should help us a bit with performance since it won't cause a reflow of the DOM. 

If this method gives you trouble though I'm fine with keeping it as `display:none` since that works too and if we don't notice a performance hit then we can probably get away with it :)

mirandawang:
Thanks so much for your contribution!!! 🙏:D

I left one comment about CSS but it's nothing huge.

rails/app/assets/stylesheets/components/card.scss
mirandawang: Thought in reference to your issue getting around the 2rem of area that's not zoomable. Is this because the HTML element is in the way of mouse clicks? If so, could we leverage `pointer-events:none` to allow the click to fall through to the element below (which should be the map)? https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events

mirandawang:
@alindeman Just wanted to check in on this PR! :)

Issue #223, opened 2019-04-30
[UI] Add spinners to import buttons


rudokemper:
Currently the import buttons on any of the admin interfaces do not show that they're working. Change the UI to where the button:

1. Is Disabled (we don't want them clicking multiple times)
2. Is Yellow (to indicate something changed)
3. Has a spinner icon (to indicate "progress")

Once the loading is done, restore the button to its previous state.

jtu0:
### Repro steps
1. Login with username and password from `seeds.rb`
1. Click on `ADMIN` button
1. _Import buttons_ are under _Speakers_, _Stories_, and _Places_

PR #233, opened 2019-05-27
[jtu+user512+alexmv] Add visual feedback to import buttons

jtu0:
When an "Import X" button is clicked, it can take a long time and users get impatient and re-click the button multiple times.

- Disable button on-click
- Change button styling and message on-click

#### Notes
- Use terrastories colors for button color change
- Only change colors for import buttons (.btn-import)
- CSS for hovering/other attributes will override, so the !important is necessary
- The terrastories codebase does *not* use bootstrap
  - Removed css classes on `<button>` that were unused references to bootstrap classes

#### Font Awesome
Added the 'font-awesome-rails' gem.  A few notes:
* FA does *not* play nicely with engines!
  * The import button is part of the admin interface, which is created with the `administrate` gem
  * `administrate` is an engine...
  * To use FA with `administrate` or any other engine, add stylesheet reference manually in the initializer(s).
* May want to consider using `font-awesome-sass` instead of `font-awesome-rails`.
  * `font-awesome-sass` has a smaller footprint
  * `font-awesome-sass` has the latest FA version, and `font-awesome-rails` does not

Resolves https://github.com/Terrastories/terrastories/issues/223

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -34,6 +34,8 @@ gem 'administrate-field-active_storage'
 gem 'webpacker', '~> 3.5'
 gem 'react-rails'
 
+gem 'font-awesome-rails', '~> 4.7'
+
 # plyr-rails gem is the integration of plyr.io javascript library for your Rails 4 and Rails 5 application.
 gem 'plyr-rails'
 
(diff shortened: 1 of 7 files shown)


jtu0:
@user512 @alexmv

PR #234, opened 2019-06-01
Fix CVE-2019-5421: Devise Vulnerability

csexton:
This changes updates to Devise 4.6.2.

> Details
>
> CVE-2019-5421
>
> Patched version: 4.6.0
>
> Devise ruby gem before 4.6.0 when the lockable module is used is vulnerable to a time-of-check time-of-use (TOCTOU) race condition due to `increment_failed_attempts` within the `Devise::Models::Lockable` class not being concurrency safe.

doctor

### DIFF
--- a/rails/Gemfile.lock
+++ b/rails/Gemfile.lock
@@ -69,7 +69,7 @@ GEM
     babel-transpiler (0.7.0)
       babel-source (>= 4.0, < 6)
       execjs (~> 2.0)
-    bcrypt (3.1.12)
+    bcrypt (3.1.13)
     better_errors (2.5.1)
       coderay (>= 1.0.0)
       erubi (>= 1.0.0)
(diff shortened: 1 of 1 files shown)


csexton:
Verified locally 🎉 Tested the following:

- ✅ Login
- ✅ Sign out
- ✅ Create Account
- ✅ Reset Password

rails/Gemfile.lock
```diff
@@ -367,4 +367,4 @@ RUBY VERSION
    ruby 2.5.1p57
 
 BUNDLED WITH
-   1.17.1
+   2.0.1
```
Bootjack: Maybe commit the change to revert this bundler vesion?

csexton: Done 😎

Issue #180, opened 2018-11-12
[CMS] Update Story importer to handle Many to Many relationships


mirandawang:
Currently, the Story importer can only import one Place, one Speaker, and one video attachment. Let's update the importing code so the CSV can contain semicolons between values for the columns on Story that have a Many to Many relationship (Place, Speaker, video).

IlinDmitry:
Hello.
Csv file should be like this?

**story_title;;speaker_id;point_id;;;;;media_filename;permission_level**
One story;;0;0;;;;video1.avi;;
Another story;;0;1;;;;video2.avi;;
One more story;;0;1;;;;video3.avi;;

I've found that in this [part](https://github.com/IlinDmitry/terrastories/blob/3e7d6f55f7c60eeaeb1044837966a95ce381e115/rails/app/models/story.rb#L8-L20) @seanmarcia used indexes 0,2,3,8,9, but indexes 1,4,5,6,7 are not used for this uploading.

mirandawang:
Hi @IlinDmitry 

Thanks for your question. If you intend to work on this issue I can remove the `help wanted` label. Please let me know!

I believe that is correct. Here is a sample CSV file that you can use for reference: 
https://docs.google.com/spreadsheets/d/1yQzVwvO2wktkgPNCH7txTxENOOcW_L2NCEuQdaMSZyE/edit?usp=sharing

mirandawang:
We have the larger model updates captured in this issue https://github.com/rubyforgood/terrastories/issues/164
Maybe you and @bperlik could coordinate since they are very related?

Sorry for the confusion! Thank you for your contributions!

bperlik:
@IlinDmitry - contact me on the Terrastories Slack channel. I will be updating the speaker model to have speaker_name, birth_date, birthplace, and photo.

rudokemper:
Some discussion on this Issue was held at Ruby by the Bay here: https://github.com/rubyforgood/terrastories/issues/9

PR #237, opened 2019-06-01
180 Update Story Model and Importer

cmhnk:
Addresses Issue #180 by:

- creating a Join Table between Stories and Speakers
- removing `speaker_id` column from `Stories` table
- updating the Stories CSV Importer to allow handle semi-colon separated media & speakers

Tested with the attached zipped CSV:
[Stories.zip](https://github.com/Terrastories/terrastories/files/3244093/Stories.zip)


cmhnk:
@mirandawang Just realized I need to do some work on the frontend to not break the map - will re-open shortly!

PR #238, opened 2019-06-01
[163] Update Speaker model to include birth_place and birthyear

### LINKED ISSUE #163: [CMS] Update the Speaker model and importer
Let's update the Speaker model to match these fields, and allow the csv importer to be able to import these fields. 

name : String
birth_year : String (or DateTime)
birthplace : Place (One to Many relationship)
photo : String (the img src)

When birth_year in 'unknown' for the importer, use 'nil'.

mirandawang:
Closes #163 

Updates to the Speaker model to include new fields - birth_place (One to Many relationship with Place) and birthyear (DateTime, in format Year)

Also includes updates to the administrate dashboard so these new fields can be added and changed. 

Importer is updated to handle the new columns too. If the CSV value is `unknown` or empty, we will set a null birthplace.

### DIFF
--- a/rails/app/models/speaker.rb
+++ b/rails/app/models/speaker.rb
@@ -1,4 +1,3 @@
-require 'pry'
 # == Schema Information ==
 #
 # Table name: speakers
--- a/rails/app/models/speaker.rb
+++ b/rails/app/models/speaker.rb
@@ -1,8 +1,24 @@
+# == Schema Information ==
+#
+# Table name: speakers
+# 
+# id          ... not null primary key
+# speaker_name... string
+# birth_year  ... datetime, nil if blank
+# birthplace  ... string, classname: place
+# photo       ... string, url to attached media
+# region ------ removed
+# community --- removed
+# created_at  ... datetime, not null
+# updated_at  ... datetime, not null
+
 class Speaker < ApplicationRecord
   has_many :speaker_stories
   has_many :stories, through: :speaker_stories
+  has_one :birthplace, class_name: Place
   has_one_attached :media
 
+  # photo
   def picture_url
     if media.attached?
       Rails.application.routes.url_helpers.rails_blob_path(media, only_path: true)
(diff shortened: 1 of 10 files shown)


rails/app/models/speaker.rb
```diff
@@ -1,7 +1,24 @@
+require 'pry'
+# == Schema Information ==
```
cmhnk: Did you mean to delete `require 'pry'`? 😸

mirandawang: Whoops, yes!

rails/app/models/speaker.rb
```diff
@@ -12,13 +29,24 @@ def picture_url
...
       end
     end
   end
 
+  def self.get_birthplace(name)
+    if name.nil? || name.downcase == 'unknown'
+      return nil
+    end
+    return Place.find_or_create_by(name: name)
+  end
```
cmhnk: NIT but you could change this to just `Place.find_or_create_by(name: name)` since Ruby methods automatically return the value of the last expression

rails/spec/models/speaker_spec.rb
```diff
@@ -1,7 +1,21 @@
 require 'rails_helper'
 
 RSpec.describe Speaker, type: :model do
-  it 'can be initialized' do
-    expect(create(:speaker)).to be_a(Speaker)
+
+  let(:place) { Place.create!(name: "Place", type_of_place: "placetype")}
+  let(:speaker) { Speaker.create!(name: 'Speaker Name', birth_year: Date.ordinal(1992), birthplace: place ) }
```
cmhnk: that's my birth year too! 😎

rails/app/models/speaker.rb
mirandawang: ah yes!!!  too much time in javascript land :P

rails/spec/models/speaker_spec.rb
mirandawang: 🤙 🤙 🤙

PR #240, opened 2019-06-01
Migrate Point model to Place model

mkmckenzie:
This PR does a few things:
1. adds a "region" column to Place
2. Migrates region data from Point to Place
3. Converts coordinates to decimal as they were in Points
4. handles relating stories to places on CSV upload of stories
5. handles creating new Places on CSV upload (adds region, lat, long)

This PR **does not** handle:
1. dropping the Points table
2. changing any references to Points in the code (incl. admin dashboards)
3. updated admin dashboards for Place/Story (if Story is needed)

**Request from reviewers:**
- Please pull down this branch and run the migrations locally and make sure the data migrated over as expected. My local data got too mangled from previous attempts and I don't have access to create a fresh db. 😢


rails/db/migrate/20190601202823_migrate_place_id_story_idto_join_table_place_story.rb
```diff
@@ -0,0 +1,11 @@
+class MigratePlaceIdStoryIdtoJoinTablePlaceStory < ActiveRecord::Migration[5.2]
+  def change
+    points = Point.all.index_by(&:place_id)
+
+     Place.find_each do |place|
+      points = Point.where(place_id: place.id).map(&.id)
```
mirandawang: Should this be
`Point.where(place_id: place.id).map(&:id)` ?

I was getting some syntax errors when trying to run the migrations.

mirandawang:
Sorry I totally missed the part where you say the PR doesn't include admin dashboard updates :sweat_smile:

PR #255, opened 2019-06-15
Drop Points model; Update other model fields

mirandawang:
This PR includes a bunch of model updates!

There's a ton of migrations... it might be easiest to test this by blowing away your DB and resetting it.

To test this PR: check out my branch and run `script/start`
When the rails container is up, run `script/console`
Inside of the rails container, run `rails db:drop` `rails db:setup`
~~ It should work!! ~~

Check the admin page to make sure things went ok.

Also try to import some data using the format from our sample tables from Google drive!

**Changes in this PR**
Firstly... Dropped the taggings and tags tables, since we don't use it.
Then, various column updates:

Story

- Included a new field: `interview_location` which is associated with Place through a foreign key id `interview_location_id`
- Has many Places (in the frontend currently we will just pass the first Place, so we don't break it. There will be a follow up PR to handle the new case of getting multiple Places per Story).

Speaker
- Updated birthyear to birthdate.
- However, the importer can only handle year format. If we decide on a new format for the importer, we will need to update the Date format string.

Place

- Has Many Stories
- All other necessary migrations were completed beforehands; thanks @mkmckenzie !

** DROPPED THE POINT MODEL! :D **

Various frontend updates to work with Places ( was a lot easier than I thought! )


rudokemper:
Would it be relatively easy to add two fields to the Stories model --

`Language` (string) (I thought @s-nicas had added this at some point)
`Interviewer` (well, this could be a string, but it could also link to Speaker, just in case a speaker was ever also the person who conducted the interview)

Or maybe they're already in the model, but just not showing up in Administrate yet?

If it's a bit of a hassle at this point, we can make a lower priority Issue out of it. It's not vital to have at all, certainly not for the MVP but just to have the models 100% complete :)

mirandawang:
@rudokemper Those fields do exist! :) I didn't realize at first. I added them to administrate so they show up. Also updated interviewer so it's not just a String, but connected to Speaker as well. If you want to reset your database, the seeds will be updated with new values, but you can also just update them manually. Let me know if you want me to add these fields to the importer too, and which columns they would be at!

rudokemper:
image

Ooh awesome! It would be these two in the spreadsheet.

mirandawang:
@rudokemper Done!

PR #263, opened 2019-08-10
262 add community field

MxOliver:
Added community field to speaker model and ran migrations. Added speaker field to the speaker dashboard, and csv importer. Added community attribute to speaker_spec.rb 

// factorybot tests from 4 months ago are failing I am not sure if they are still relevant or if it is something with the spec_helper configuration? Wanted to get this PR in but will update with test changes when I can figure out what is happening

### DIFF
--- a/rails/app/controllers/speakers_controller.rb
+++ b/rails/app/controllers/speakers_controller.rb
@@ -41,6 +41,6 @@ class SpeakersController < ApplicationController
   private
 
   def speaker_params
-    params.require(:speaker).permit(:name, :media)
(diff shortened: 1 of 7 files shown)


mirandawang:
Sorry for the SUPER late response @MxOliver 

We are changing our DB around a little bit to make development easier. It might be Postgres now instead of Mysql, so I am gonna wait on approving this PR until we have that confirmed.

Thanks for your patience! <3 We really appreciate the contribution!!!

mirandawang:
Reopened here: https://github.com/Terrastories/terrastories/pull/272

PR #265, opened 2019-08-22
reset scrollbar to top on select

MxOliver:
In root App component added ref.current.scrollIntoView function on the handleStoryClick method, then passed the scrollRef to StoryList and attached it to the first story based on index. 

(Sorry for the previous closed pull request, forgot to turn off my prettier auto format and my text editor changed the format of a million diff files which were then added to the commit and rebasing was a mess)

### DIFF
--- a/rails/app/javascript/components/App.jsx
+++ b/rails/app/javascript/components/App.jsx
@@ -15,6 +15,7 @@ class App extends Component {
       activePoint: null,
       activeStory: null
     }
+    this.scrollRef = React.createRef();
   }
 
   static propTypes = {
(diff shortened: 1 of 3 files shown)


mirandawang:
image
thanks for another contribution!! <3

i was able to rebase your branch on top of master with no conflicts, and it ran well!
however im not sure im seeing the behavior we expect - the scrollbar seems to be staying in its current position when i filter or reset the filter

Oct-06-2019 15-20-16

MxOliver:
Hmmm let me look into it some more

MxOliver:
Ahh okay so I was only resetting the scroll bar when you select a person/story not with the drop down filters. I think I can just pass the ref to that filter but I wanted to clarify. Should it still reset to top when you click on a person, or only when the drop down filters change?

mirandawang:
@MxOliver  Sorry for the confusion! Yeah, it should stay where it is when you select a story, but reset to the top when you apply any of the filters :)

mirandawang:
closed since we have this(: https://github.com/Terrastories/terrastories/pulls?q=is%3Apr+is%3Aclosed

PR #269, opened 2019-09-18
Simple non-docker development enviroment

csexton:
WARNING: This may break docker installs, but it might not. Please test before deploying!

Switch to use sqlite3 for the database. This means we don't have to have a SQL DB server running at all. Long term I think the production deploys on NUCs and other field hardware could use this. But the migration will need some work.

Add dotenv to load the .env file when running locally.

Minor changes to how webpack is loaded. Mostly pulled from the rails defaults.


csexton:
This will need new instructions for running locally. Basically it can all be down w/o docker. We can get docker working in the future, but I'd go with the non-docker approach first (but I am biased there)

Steps for running in development:

- Install Ruby 2.5.1
- Clone Repo
- `bundle install`
- `rails db:setup`

Then to run it, run each in a terminal tab:

```
./bin/rails server
./bin/webpack-dev-server
```

mirandawang:
is this still needed?
closing for now and thank you for the work!

csexton:
Do you still want to be able to run w/o containers? These changes have a few things you may want to cherry-pick into master if so. If not--you are all good!

PR #273, opened 2019-09-29
Update SETUP.md

MxOliver:
Probably not super important but 3, and 4 from the table of contents had an extra space that prevented them from being links


mirandawang:
Thank you!!!

Issue #224, opened 2019-04-30
[Import/export] Ensure that duplicate items cannot be added to the database


rudokemper:
Currently when the `Stories` and `Places` CSV file are imported, if it's run again it will add duplicate records.

Let's add validations to the `Stories` and `Places` models to ensure that duplicates cannot be created, and that the CSV importer can do idempotent imports. 

Validations are already in place for the `Speakers` importer, the logic for which can be used to update the `Places` and `Stories` importers as well.

rgenchev:
Can I help with adding validations, indexes and tests? I could do that in a new PR.

mirandawang:
Some work has started here, but can be picked up by whoever is available! https://github.com/Terrastories/terrastories/pull/274

PR #274, opened 2019-10-03
Prevent duplication story during csv import

mfcruz:
Solution for #224 
Uniqueness is based on story title.
Adding model validation will also prevent further duplication of entry.

### DIFF
--- a/app/models/story.rb
+++ b/app/models/story.rb
@@ -8,32 +8,32 @@ class Story < ApplicationRecord
 
   def self.import_csv(filename)
     CSV.parse(filename, headers: true) do |row|
-      story = Story.new(title: row[0])
-      story.desc = row[1]
-      # add each spaker to the story, split by comma
-      speaker_names = row[2]
-      speaker_names.split(',').each do |speaker|
-        story.speakers << Speaker.find_or_create_by(name: speaker.strip)
+      Story.where(title: row[0].first_or_create do |story|
+        story.desc = row[1]
+        # add each spaker to the story, split by comma
+        speaker_names = row[2]
+        speaker_names.split(',').each do |speaker|
+          story.speakers << Speaker.find_or_create_by(name: speaker.strip)
+        end
+        # add each place to the story, split by comma
+        place_names = row[3]
+        place_names.split(',').each do |place|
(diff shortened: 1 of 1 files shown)


app/models/story.rb
```diff
@@ -8,32 +8,32 @@ class Story < ApplicationRecord
...
+        # Add interviewer
+        story.interviewer = row[6].blank? ? nil : Speaker.find_or_create_by(name: row[6])
+        # Add Language
+        story.language = row[7].blank? ? nil : row[7]
+        if row[8] && File.exist?(Rails.root.join('media', row[8]))
+          file = File.open(Rails.root.join('media',row[8]))
+          story.media.attach(io: file, filename: row[8])
+        end
+        story.permission_level = row[9].blank? ? "anonymous" : "user_only"
       end
```
mirandawang: Missing a `)` parenthesis here :)

rudokemper:
I do think it would be good to distinguish stories on a more substantial basis than just story title alone.

To illustrate, imagine that there are two stories titled "Oral History of St. Petersburg". One is about St. Petersburg, Florida and the other about St. Petersburg, Russia. Distinct stories and places, but same title.

In all likelihood, Terrastories won't be used to map oral histories about Central Florida and Russia in the same instance, but in indigenous territories there are distinct places with the same name too.

Or imagine a case where Terrastories is being used to map family histories within a community. And multiple people want to use the title "My Grandma's History" or something like that. I don't think we should expect users to ensure their stories have unique titles.

But then, I'm not sure what could guarantee uniqueness for Stories? My thinking was video filename, but what if there are multiple `MyGrandmasHistory.mp4`s? Same issue.

What if it just checks for _all_ of the fields in the story model? I think that would be robust enough for sure. And we can just make a note somewhere in the documentation for users to avoid across-the-board duplication of field content in the spreadsheets.

Thoughts?

midhunkrishna:
@rudokemper may be title, locations and content?

rudokemper:
Sounds good!

mirandawang:
Hey @mfcruz , are you still able to work on this? :)

mirandawang:
Closing due to inactivity

PR #277, opened 2019-10-05
Clean and re-configure Rspec with SimpleCov gem for test coverage

MxOliver:
I removed the spec directory, reconfigured rspec and set up a simple test of the speaker model to make sure things were running smoothly, then added simplecov for test coverage which is also running well.


MxOliver:
I think I fixed the conflicts

PR #280, opened 2019-10-06
Initial testing documentation

MxOliver:
Sorry for the double PR accidentally had some unrelated commits on the other one so I closed it. Added some testing documentation to the set up guide

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -15,12 +15,14 @@ This instructions are for Mapbox Community Days and Hacktoberfest, which everyon
 
 6. [Adding languages to Terrastories](#adding-languages-to-terrastories)
 
+7. [Running and Writing Tests](#testing)
+
 ## Docker Prerequisites
 
 Install docker. On linux, you may have to install docker-compose separately.
 
- - https://docs.docker.com/install/
- - https://docs.docker.com/compose/install/
+- https://docs.docker.com/install/
+- https://docs.docker.com/compose/install/
 
 On Windows, all terminal docker commands need to be run using Windows PowerShell, not Command Prompt.
 PowerShell comes with Windows.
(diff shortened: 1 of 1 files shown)


SETUP.md
```diff
@@ -109,3 +111,32 @@ For the `devise` and `administrate` files, there might be available translations
...
+
+## Testing
+
+The terrastories application is currently using rspec for testing, to run existing tests make sure you are within
+the docker container (run `$ docker exec terrastories_web_1 /bin/bash`) then run
+
+```
+$ docker-compose run web scripts/wait-for-it.sh db:5432 -- "rspec spec"
+```
+
```
mirandawang: I had trouble running this when I was on a new branch and needed to install gems. The way that I was able to run it was to do:
`docker exec -it terrastories_web_1 /bin/bas`
which put me into the rails container command line and then directly run `rspec spec` The other command was fine when I didn't need any new gems installed. What do you think of adding this as another way of running the tests in the container?

kalimar:
As a permanent solution I don't think we want to have the practice of entering the container. Instead, we should have our conventional rails scripts wrapping these commands as well as using entrypoint.sh to simplify the docker commands themselves. I think this issue is probably more indicative of a problem in how I setup the development environment.

mirandawang:
@kalimar How should we run rails commands while we don't have the wrapper scripts?
Should everything be thru `docker-compose run web scripts/wait-for-it.sh db:5432 -- "<<rails command>>"` ?

mirandawang:
Actually, I'm not sure how I was able to stay in the rails container when the gems weren't installed. 
I think it was because it was already running before I switched branches?

But when I tried to restart the container, `docker-compose up` would exit upon the gem error. I couldn't get back into the container to run `bundle`. I tried running `bundle` with `docker-compose run web scripts/wait-for-it.sh db:5432 -- "bundle install"` and it says it installs gems correctly but I still can't start the container without the gem error.

mirandawang:
Ok nvm, we have to do a `docker-compose build` every time gems or npm packages change!

kalimar:
Yeah, sorry my comment wasn't really complete. Docker-compose build runs bundle install and copies everything into the container it's creating. When you enter the container and make changes those will dissappear when you close it. So if there are gems being added, the PR should include the new gemfile.lock and/or package.json - no need for Node_modules. Too big.

SETUP.md
```diff
@@ -109,3 +111,32 @@ For the `devise` and `administrate` files, there might be available translations
...
 Once you are done, the language should be available the next time you start Terrastories.
+
+## Testing
+
+The terrastories application is currently using rspec for testing, to run existing tests make sure you are within
+the docker container (run `$ docker exec terrastories_web_1 /bin/bash`) then run
+
+```
+$ docker-compose run web scripts/wait-for-it.sh db:5432 -- "rspec spec"
+```
```
mirandawang: I think this needs to be updated to running:
`docker exec -it terrastories_web_1 /bin/bash` then `rspec spec` inside the container 

Otherwise rest of it looks good!!

MxOliver: Good point, will do that tomorrow!

SETUP.md
```diff
@@ -109,3 +111,32 @@ For the `devise` and `administrate` files, there might be available translations
...
 
 Once you are done, the language should be available the next time you start Terrastories.
+
+## Testing
+
+The terrastories application is currently using rspec for testing, to run existing tests make sure you are within
+the docker container (run `$ docker exec terrastories_web_1 /bin/bash`) then run
+
+```
+$ docker-compose run web scripts/wait-for-it.sh db:5432 -- "rspec spec"
```
kalimar: Tests will run with just `docker-compose run web scripts/wait-for-it.sh db:5432 -- "rspec spec". There's no need to `bash` into the container.

kalimar: In the near future we'll add either a script or an `entrypoint.sh` to reduce verbiage associated with `scripts/wait-for-it.sh`

MxOliver: Do we have that entrypoint script now? I for some reason just saw these comments as I was going through PRs

mirandawang:
@kalimar @MxOliver Should we update this or close this PR from a while ago?

MxOliver:
I think we should close it

Issue #287, opened 2019-10-06
[Test] Add test for the story model


MxOliver:
Add tests for the story model (`app/models/story.rb`)

midhun-at-vineti:
@MxOliver @kalimar Picking this up.

midhunkrishna:
@kalimar @MxOliver Picking this up. 👍

midhunkrishna:
@MxOliver we can close this issue?

PR #295, opened 2019-10-08
287 add test coverage for story model

midhunkrishna:


midhunkrishna:
@MxOliver Fixes https://github.com/Terrastories/terrastories/issues/287

midhun-at-vineti:
@MxOliver @kalimar This PR needs review.

MxOliver:
I can review this tomorrow when I can run it on my laptop

app/models/story.rb
```diff
@@ -1,41 +1,22 @@
 class Story < ApplicationRecord
+  MEDIA_PATH = Rails.env.test? ? 'spec/fixtures/media' : 'media'
```
namel3ss: The media path is used by other models as well. I was thinking, why not add it as a custom Rails configuration variable, and set it in the different environment configs. Like:
```
# application.rb
Rails.application.configure do
  config.media_path = 'media'
end
```
overwrite it in `test.rb` and then just call:
```
Rails.configuration.media_path
```

midhunkrishna: @namel3ss Can be done. I thought of doing this initially, but then thought to keep it the was it is currently until @MxOliver or @kalimar takes a pass at reviewing this PR. Also, thank you for waiting on this PR for your #300

MxOliver: I think that sounds like a good idea, @kalimar thoughts? Also I've approved this PR so you can go ahead and merge so that @namel3ss can rebase

mirandawang:
Hey there! Thanks so much for the contribution!!

Looks good to me as well, we can merge it in once Gemfile.lock conflicts are fixed @midhunkrishna

midhunkrishna:
@mirandawang  @MxOliver Please merge 👍

MxOliver:
This PR is actually already closed, unless you mean the ticket? I can check and make sure the ticket is marked as done

Issue #259, opened 2019-07-03
[Map] Closing a marker popup should not reset the map zoom/bounds


rudokemper:
This is an easy fix: when you close a popup, it should not reset the map view to home.

(If you're looking at a bunch of clustered markers close by each other, it's a little annoying to have to keep zooming in to examine the others.)

rudokemper:
Studying `Map.jsx` a little, I can tell that both the **filter** map reset as well as the **close-marker** map reset (and just about any other action) are handled by the `else` in this conditional;

```    if (this.props.pointCoords.length > 0) {
      if (this.map) {
        this.map.panTo(this.props.pointCoords);
      }
      return;
    } else {
      if (this.map) {
        this.resetMapToCenter();
      }
    }
  } 
```

But I'm not sure how to take it from here. Maybe a specific conditional for if-marker-popup-is-closed, `doNothing`; or alternatively, a default action of `doNothing` except for when filters are cleared, in which case, `resetMapToCenter`.

leosoaivan:
May I attempt this one?

kalimar:
Yeah definitely! Please let us know if you need any clarification or maybe a screen video.

leosoaivan:
I posted in the slack, but perhaps issues related to this ticket belong here :)

Ran into a pending migration error after `docker-compose up`:

```
Starting terrastories_db_1 ... done
Starting terrastories_web_1 ... done
Attaching to terrastories_db_1, terrastories_web_1
db_1   | 2019-10-08 11:08:40.449 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
web_1  | wait-for-it.sh: waiting 15 seconds for db:5432
db_1   | 2019-10-08 11:08:40.449 UTC [1] LOG:  listening on IPv6 address "::", port 5432
web_1  | wait-for-it.sh: db:5432 is available after 0 seconds
db_1   | 2019-10-08 11:08:40.452 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
db_1   | 2019-10-08 11:08:40.465 UTC [25] LOG:  database system was shut down at 2019-10-08 11:08:36 UTC
db_1   | 2019-10-08 11:08:40.470 UTC [1] LOG:  database system is ready to accept connections
db_1   | 2019-10-08 11:08:41.042 UTC [32] LOG:  incomplete startup packet
web_1  | => Booting Puma
[...]

PR #298, opened 2019-10-09
[259] Maintain map view on marker popup close

leosoaivan:
This PR addresses #259, where closing a popup marker caused the map to re-center.

After drilling down a bit, and assuming my understanding is correct 😅, `resetStoriesAndMap` in App.jsx sets `pointCoords` to an empty array, thereby triggering `this.resetMapToCenter`.

Because `this.resetMapToCenter` is still needed for instances where the filters are cleared, it seemed to me that a quick way forward was to create a separate handler for closing the popup that *doesn't* clear `pointCoords`.

Hopefully haven't butchered anything :).


mirandawang:
Thank you so much for your contribution!!!

This is looking great. The map stays in the same spot once we close a popup!
But we also want the Stories in the sidebar to no longer be filtered to that place (since it's technically no longer selected). Can we add that functionality back in? The sidebar should repopulate back with all of the stories when no place is selected.

leosoaivan:
@mirandawang thank you for the review!

Sorry about the filtering, will take a look this week ASAP!

leosoaivan:
@mirandawang I noticed something while re-implementing the stories, and this is true in `master` as well, but when closing the pop-ups, the filter doesn't get cleared (unless I'm mistaken).

Given that closing a pop-up won't re-center the map, it might be confusing if you use the filter to drill down to a specific area, click on story, close the pop-up, and see the sidebar re-populated with all the stories while the filter is activated.

If we want to re-populate the stories, it might make sense to maybe also implement functionality to clear the filter? Maybe another ticket, or maybe it already exists and I missed it? Anyway, just a thought.

mirandawang:
@leosoaivan Thanks for flagging! I didn't realize that the filters don't get reset when you close a pop up. I believe we want this behavior, cc @rudokemper is that true? The sidebar currently will display only the stories associated with a specific place if that place is selected on the map. If we close the marker, should the stories also get reset (without clicking on another place)? 

If this is extra behavior we would like, it can totally be a separate ticket from this one.

Then we can merge this work in! Thanks @leosoaivan ! :)

rudokemper:
I agree with both of you @leosoaivan and @mirandawang; let's reset the filters as well. Thanks for noticing @leosoaivan and your work on this important issue!

leosoaivan:
@mirandawang @rudokemper Just wanted to check-in re: an approach to resetting the filter.

It seems like there's two ways to handle this. First, using `componentWillReceiveProps` or `getDerivedStateFromProps` to handle reseting Filter.jsx from App.jsx. Second, is fully controlling Filter.jsx's state from App.jsx.

The first approach is definitely the easiest, but my understanding is that it's prone to anti-patterns and unintended side effects. The second approach is likely the most [correct](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions). It also makes more sense to me, since it seems we are lifting so much state from Filter.jsx back up to App.jsx. It would require a re-write, but I think we rewrite Filter.jsx into a functional component, which would make it easier to implement hooks in the future.

Any thoughts, preferences, or corrections? Much appreciated :).

kalimar:
@leosoaivan thanks again for noticing the filter issue.  I also agree with your comments here git checkout -b leosoaivan-259-closing-marker-popup master
git pull https://github.com/leosoaivan/terrastories.git 259-closing-marker-popup
 I'll reference all of this in the new ticket.

@mirandawang I'll go ahead and create a new ticket for handling the filter, let's just review this one and merge it. 

@leosoaivan - I tried running your branch locally but I think at this point it might be a bit too different from the current master, I was unable to run it locally. Take a look and see what needs to be changed. It might even be easier to save your changes elsewhere, start a fresh branch and add your updates? Let me know if it's too much work. I'm happy to help you get unblocked so we can merge the work. I really appreciate all your help.

kalimar:
Created a ticket https://github.com/Terrastories/terrastories/issues/364

leosoaivan:
@kalimar I merged master into my branch, hopefully that's resolved any issues. If not, I can re-create my work hah!

mirandawang:
@leosoaivan Due to the age of this PR, it may be hard to reconcile w/ the current master state. Could you possibly open a new one? Thank you!! (Sorry for the super late response; totally ok if you no longer have the time to work on this!)

leosoaivan:
@mirandawang I'll give it a go this weekend.

PR #331, opened 2019-10-13
[#296] Add bundle and migrate to docker-compose

### LINKED ISSUE #296: Execute bundle and rails db:migrate on docker-compose up
When there are new gems added to the app or pending migrations, it is a lot more easier to just stop the composer and up it again. We can add the `bundle` and `rails db:migrare` commands to the composer. 

Reason for this suggestion is the following issue:

I stopped the container and there was a new gem. I couldn’t run the container with `docker-compose up` anymore because rails -s was failing.

namel3ss:
Execute `bundle` and `rails db:migrate` on `docker-compose up`. This
helps to manage the dev environment easier when there are new gems or
migrations added to the project. The developer only needs to stop the
running container and run `docker-compose up` to get everything up to
speed.

Fixes #296

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,11 @@ services:
       - db
     ports:
       - 3000:3000
-    command: ["./scripts/wait-for-it.sh", "db:5432", "--", "./scripts/start_rails.sh"]
+    command: [
+      "./scripts/wait-for-it.sh",
+      "db:5432",
+      "--",
+      "./scripts/start_rails.sh"
+    ]
     volumes:
       - .:/opt/terrastories:cached
--- a/scripts/start_rails.sh
+++ b/scripts/start_rails.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-rm -f tmp/pids/server.pid && bundle exec rails s -b 0.0.0.0
+rm -f tmp/pids/server.pid && bundle && rails db:migrate && bundle exec rails s -b 0.0.0.0


kalimar:
In principal I don't think this is really a best practice, because updating gems and migrations should be part of a build process. That said I want development to be as painless as possible. I'd like to 1. Have folks try it locally 2. Folks can chime in their feelings about it. Then merge. 

Thoughts?

namel3ss:
I understand your point of view. For me, so far, running `bundle` and `rails db:migrate` has been sort of a regular ritual when working on projects with a team or a lot of people. From point of view of deployment it is 100% a build step. From point of development environment, I think the build should contain stuff that don't change often when installed/configured. In case of the gems and migrations, we usually have a lot of changes - updated gem versions, new models, new columns, new indexes, migrating data, etc. In my practice so far, almost after every rebase from master, I need to run these. From that standing point, it is easier for me to just automate these two commands when I spin up my dev environment (be it container or locally starte Puma), and don't have the overhead to build the dev environment every time there is a new change in the `Gemfile` or the DB schema. Keep in mind that I am using dockerized development environment only for this one and my current job project. Haven't used that approach in the other projects I have worked on so far. In my current project, I find it easy and useful to have these 2 commands on `docker-compose up`.

mirandawang:
@kalimar @namel3ss 

Do we want to not include `rails db:migrate` in the `docker-compose up` command?
Or does this need further discussion? in which case I'd like to close this PR for now and discuss in Slack or an issue labelled discussion.

Thanks!

kayhide:
Rails comes with setup script of `bin/setup`, which does the regular ritual like bundle, db setup, etc...

I opened a PR at #356 which reflects my suggestion of how to setup using that script.

If the setup is easy enough, like just one line of command, I don't think we want it in the docker up command.

mirandawang:
@kayhide Thank you for the response, and sorry for this delayed one! I will take a look at your PR shortly once I get to my laptop w/ terrastories set up in it.

Closing this issue due to age & inactivity

Issue #284, opened 2019-10-06
[Test] Add test for the stories_controller


MxOliver:
Add tests for the stories_controller (`app/controllers/stories_controller.rb`)

rgenchev:
Can I work on this? :)

midhunkrishna:
@rgenchev Are you working on this?

rgenchev:
Yup :)

rgenchev:
@midhunkrishna, in order to test the controller actions, we need views. Should I create empty ones in this PR?

rgenchev:
I created https://github.com/Terrastories/terrastories/pull/333 that should resolve this. I'd create a new issue for integrating https://github.com/paulelliott/fabrication. It'd really help when writing tests.

PR #333, opened 2019-10-13
[#284] Add tests for StoriesController

rgenchev:
Closes https://github.com/Terrastories/terrastories/issues/284

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,7 +82,6 @@ group :test do
   gem 'chromedriver-helper'
   # Factories for creating database entities for testing
   gem 'factory_bot_rails'
-  gem 'rails-controller-testing'
   gem 'simplecov', require: false
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,7 @@ group :test do
   gem 'chromedriver-helper'
   # Factories for creating database entities for testing
   gem 'factory_bot_rails'
+  gem 'rails-controller-testing'
   gem 'simplecov', require: false
 end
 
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -82,6 +82,7 @@ group :test do
   gem 'chromedriver-helper'
   # Factories for creating database entities for testing
   gem 'factory_bot_rails'
+  gem 'rails-controller-testing'
```
namel3ss: Since Rails 5 it is recommended to use the built in RSpec request tests, and not use the `rails-controller-testing` gem.

mirandawang:
@rgenchev How do you feel about the above ^ suggestions? I think it would be best to avoid introducing new dependencies if we don't need them!

rgenchev:
> @rgenchev How do you feel about the above ^ suggestions? I think it would be best to avoid introducing new dependencies if we don't need them!

@mirandawang , I totally agree. I'll push the changes tonight. :)

spec/controllers/stories_controller_spec.rb
```diff
@@ -1,5 +1,162 @@
...
+    end
+
+    it 'should redirect to :show' do
+      expect(subject).to redirect_to(story_path(assigns(:story)))
+    end
+  end
+
+  private
+
+  def fabricate_stories
```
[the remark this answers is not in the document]
rgenchev: Done! :)

Gemfile
rgenchev: I reverted the commit.

Issue #319, opened 2019-10-12
Setup Istanbul Test Coverage tool


kalimar:
https://istanbul.js.org/

GQuirino:
Can i take a look into this issue?

kalimar:
@GQuirino we made some requirement changes but feel free to continue working on this. https://github.com/Terrastories/terrastories/milestone/7 - You can configure Istanbul within Jest. I tried a few different things but no luck - test coverage reports came up empty. I'm pretty sure it's one or two of the configuration options in `package.json`

PR #334, opened 2019-10-15
Feature/319-Setup-Instambul-Test-Coverage

GQuirino:
Issue: #319 
Following guideline of https://istanbul.js.org/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,4 @@ yarn-debug.log*
 /node_modules
 yarn-debug.log*
 .yarn-integrity
+.nyc_output/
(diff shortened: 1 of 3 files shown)


kalimar:
Hey @GQuirino , I'm very sorry but  the tickets I created are not very clear. It turns out there's more investigating to do. First, https://jestjs.io has [Istanbul included](https://istanbul.js.org/docs/tutorials/jest/) so we can set that all up there. Second, it looks like there are some legacy configurations in our `package.json` file that are causing problems. I tried setting up the coverage reporting tool following some of the suggestions [here](https://github.com/facebook/jest/blob/master/docs/Configuration.md#collectcoveragefrom-array) and have been unsuccessful. If you have some time and want to also take a look, that would be great.  I'm going to add additional commentary and requirements to the original ticket and close this PR for now. Let me know if you have any questions.

PR #339, opened 2019-10-18
[#226] Frame map view around filtered stories

### LINKED ISSUE #226: [Map] Zoom to maximum extent of filtered points
In addition to only showing the points that match the stories shown in the sidebar, the map also should zoom to the maximum geographical extent of the points shown on the map, when selecting a filter.

Some discussion about this issue took place at Ruby by the Bay 2019, here: https://github.com/rubyforgood/terrastories/issues/37

sansumbrella:
When filtering stories in the sidebar, the map now zooms and pans to the region containing those stories. This makes them better fill the available screen space.

## What changed
Replaced `pointCoords` with `framedView` object.
`framedView` allows for specifying either bounds to fit or other map properties to easeTo.
Handle changes to `framedView` prop in map to animate the map view.

## How to test
Launch the map view and confirm:
1) clicking on a story still focuses the story
2) filtering the set of stories in the sidebar reframes the map around those stories
3) removing the filter resets the map view

Closes #226

### DIFF
--- a/app/javascript/components/App.jsx
+++ b/app/javascript/components/App.jsx
@@ -9,7 +9,8 @@ class App extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      pointCoords: [],
+      framedView: null, // store information about how view should be laid out
+      pointCoords: [], // TODO: replace usage with framedView
       points: {},
       stories: this.props.stories,
       activePoint: null,
@@ -106,7 +107,14 @@ class App extends Component {
     }
     if (filteredStories) {
       const filteredPoints = this.getPointsFromStories(filteredStories);
-      this.setState({ stories: filteredStories, points: filteredPoints, pointCoords: [] });
+
+      const framedView = {
+        center: [-108, 38.5],
+        zoom: 8.5,
+        pitch: 0,
+        bearing: 0
+      };
+      this.setState({ stories: filteredStories, points: filteredPoints, framedView });
     }
   }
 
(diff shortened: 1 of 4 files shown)


mirandawang:
Thanks so much for this work @sansumbrella !!!

I checked out your branch, but am noticing some funky behavior. It appears that the bounding box that is calculated is much too large. Or it's possible there's a bug when there only exists one place to the Stories?

I'm also wondering if there's a way to calculate the bounding box without needing an extra library @turf. If we can avoid adding new dependencies, that would be great!

kalimar:
I'll just speak on behalf of the inclusion of turf here. For context I'm the one who's been advocating for extra care when adding libraries. In this case it's a submodule of turf for performing one calculation. It's tiny and makes life much easier. I'd vouche for this addition. We'll probably be adding more turf methods down the line anyway. 😊

mirandawang:
Cool I am down to use @Turf!

Can we double check on the calculation of the bounding box ? Here's what I'm seeing:
ezgif-6-da5f16a3f6f4

kalimar:
@mirandawang can you clarify what you are expecting to see in that gif?

For what it's worth, the side bar poses a challenge for any map camera motion. The map area actually includes the side bar so sometimes the behavior isn't always what you might expect. Do we want to change our default UX to have the sidebar closed? Otherwise if the bar is to always be open, then I think we should have the map be smaller at all times on desktop.

PR #357, opened 2019-10-25
Create Japanese translation

kayhide:
Created Japanese translation :jp:

こんにちは！
image


rudokemper:
This is so cool! Thanks @kayhide :)

PR #359, opened 2019-10-25
Add i18n gems

kayhide:
Added i18n related gems.
They save us by failing translation of common rails messages like validation error messages.

Here is an example:

Before)
image

After)
image


kayhide:
And after merging this, we can drop `locales/xx/devise.xx.yml` files if we are happy with the translations which come from `devise-i18n` gem.

PR #361, opened 2019-10-25
Fix users search on admin page

### LINKED ISSUE #358: Users search at admin page is broken
When trying to search users at the admin page, it raises an error.

Searching users with `xxxx` gives:

```rb
ActiveRecord::StatementInvalid at /admin
PG::UndefinedColumn: ERROR:  column users.password does not exist
LINE 1: ...email" AS CHAR(256))) LIKE '%xxxx%' OR LOWER(CAST("users"."p...
                                                             ^
: SELECT  "users".* FROM "users" WHERE (LOWER(CAST("users"."email" AS CHAR(256))) LIKE '%xxxx%' OR LOWER(CAST("users"."password" AS CHAR(256))) LIKE '%xxxx%') LIMIT $1 OFFSET $2
```

kayhide:
Excluded `password` field from searching targets.

Fix #358

### DIFF
--- a/app/dashboards/user_dashboard.rb
+++ b/app/dashboards/user_dashboard.rb
@@ -11,7 +11,7 @@ class UserDashboard < Administrate::BaseDashboard
     id: Field::Number,
     email: Field::String,
     role: EnumField,
-    password: Field::String,
+    password: Field::String.with_options(searchable: false),
     created_at: Field::DateTime,
     updated_at: Field::DateTime,
   }.freeze
--- a/app/dashboards/user_dashboard.rb
+++ b/app/dashboards/user_dashboard.rb
@@ -11,7 +11,7 @@ class UserDashboard < Administrate::BaseDashboard
     id: Field::Number,
     email: Field::String,
     role: EnumField,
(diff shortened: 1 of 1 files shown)


app/dashboards/user_dashboard.rb
```diff
@@ -11,7 +11,7 @@ class UserDashboard < Administrate::BaseDashboard
     id: Field::Number,
     email: Field::String,
     role: EnumField,
-    password: Field::String,
+    password: Field::String.with_options(searchable: false),
```
lauramosher: Instead of specifying searchable false here, lets take advantage of [Administrate's `Field::Password` type](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/password.rb) which will do the searchable false configuration by default and be more descriptive of the field attribute type.

```suggestion
    password: Field::Password,
```

kayhide: Hi @lauramosher !
Yeah, that looks better, I will take it!

lauramosher:
Thank you so much for making those updates!!

Issue #364, opened 2019-10-26
[Map] Reset the search filters 


kalimar:
@leosoaivan noticed that search filters don't clear when you close a pop-up. 

We want to rewrite `filter.jsx` to be a functional component and follow [one of these patterns](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#preferred-solutions) of which we can discuss the pros and cons in the comments below.

neckenth:
@kalimar I'd like to give this a shot. Can you assign me?

kalimar:
Woot! Go for it!

neckenth:
@kalimar I just started this - I was able to change `Filter` from a class-based component to a functional component, as requested.

Here's a link to this component on my fork for your reference: https://github.com/neckenth/terrastories/blob/neckenth/364/clear-filters-upon-close/app/javascript/components/Filter.jsx

Would love to discuss your recommendation for the pattern you'd like to follow for clearing the filters.

matthewcm:
I'll get on this issue :)

neckenth:
@kalimar I gave this a shot and opened a PR for my solution. If you want to handle component state in a different way, please let me know.

kalimar:
@neckenth thank you. Sorry I couldn't be a little more proactive there. I'll review the PR today. Your help is much appreciated.

PR #369, opened 2019-10-29
# 364 - clear filters upon closing Card

neckenth:
#364 
## Problem:
filters_bad

When using the filters and viewing a Card on the map, if you close the Card, the filters don't get reset.

This is because the category and item filters were local to the `Filters` component, whereas the function that gets invoked when closing the card is much higher up in the state hierarchy, having no direct impact on the `Filters` local state.

## Solution
filters_good

Hoist `Filters` state higher up into the `App` component, and instead pass the values down as `props`.

### Changes
* make the `Filters` component functional, as requested
* add to `App` state the following to be passed all the way down to the `Filters` component
    * `filterCategory`
    * `filterItem`
    * `itemOptions`
* add handler functions to `App` component to handle changes to the category and item filters, and pass these down to the `Filter` component as well
    * `handleFilterCategoryChange()`
    * `handleFilterItemChange()`
* update `props` passing through `Card` and `StoryList` components

### DIFF
--- a/app/javascript/components/Filter.jsx
+++ b/app/javascript/components/Filter.jsx
@@ -6,38 +6,9 @@ const Filter = props => {
   let DEFAULT_CATEGORY_PLACEHOLDER = I18n.t("select_category");
   let DEFAULT_ITEM_PLACEHOLDER = I18n.t("select_option");
 
-  const [itemOptions, setItemOptions] = useState([]);
-  const [categorySelectValue, setCategorySelectValue] = useState(
-    DEFAULT_CATEGORY_PLACEHOLDER
-  );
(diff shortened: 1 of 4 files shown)


app/javascript/components/Filter.jsx
```diff
@@ -1,100 +1,61 @@
-import React, { Component } from 'react';
-import Select from 'react-select';
-import PropTypes from 'prop-types';
+import React, { useState } from "react";
```
MxOliver: nice, thanks for updating this to a functional style component

mirandawang:
@neckenth I've rebased your work on top of master and verified the changes are still working great! Thank you so much for your contribution!!! <3

Issue #266, opened 2019-09-17
[UI] Add a visual cue to distinguish between anonymous and restricted content


mirandawang:
If you are logged in as a user, you are able to see more content than someone anonymous.

Let's add some kind of visual marker to distinguish between anonymous and restricted stories and points.

@rudokemper Do you have any ideas as to what this visual marker could be?

Whoever takes on this ticket can feel free to input their design opinion as well! :)

rudokemper:
Perhaps a lock in an unlocked position? That's something most communities everywhere have familiarity with :)

Issue #168, opened 2018-11-05
[UI] Add visual cue to distinguish between anonymous and restricted content


rudokemper:
We should add some kind of a visual (like a small lock image in the stories) to show logged in users which stories are restricted.

PR #374, opened 2019-10-31
Add a small lock emoji next to story title if the story permission_level is not anonymous

albertchae:
- I chose to combine user_only and editor_only permission_levels into
one "restricted" value since the UI doesn't care about the difference
between the two at the moment. 
I can change this to all 3 values of the enum if we think it's a good idea

- I also briefly tried using the font-awesome lock instead of the emoji, but couldn't figure out how to get it to render...

Looks like there are 2 issues for this?
https://github.com/Terrastories/terrastories/issues/168
https://github.com/Terrastories/terrastories/issues/266

Screenshot of what the change looks like:
Screen Shot 2019-10-31 at 3 43 55 AM

### DIFF
--- a/app/javascript/components/StoryList.jsx
+++ b/app/javascript/components/StoryList.jsx
@@ -103,7 +103,7 @@ class StoryList extends Component {
             }
           </div>
           <div className="container">
-            <h6 className="title">{story.title}</h6>
+            <h6 className="title">{ story.permission_level === "restricted" && "🔒" }{story.title}</h6>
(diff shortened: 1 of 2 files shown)


rudokemper:
👋 heya Albert! Hope all is well! Thanks so much for this very cool PR 😄

albertchae:
👋 good to hear from you Rudo! Hope all is well with you too. It's been great to see all the activity around terrastories!

PR #375, opened 2019-11-06
[183] Update default paragraph text color

leosoaivan:
This PR adds the color `#474747` as `$default-text` and applies it to:
- `@mixin ui-font`
- `<h6>` tags
- React selector filters and filter contents
- Mapbox pop-up text


leosoaivan:
@MxOliver Thanks for the feedback, will get on it!

leosoaivan:
@MxOliver Added some changes. Could not find instances of h3 tags to visually inspect, but did find h6's. Let me know if they should be edited regardless. Thanks!

MxOliver:
If you could add it to all the text tags in this file (so p, h1, h2, h3, h4, h5) `app/assets/stylesheets/core/_typography.scss` that would be great, because they all currently default to black and we want them to default to the new default text color. That way, even though we may not have any h3 elements in use somewhere yet, if we add a h3 tag somewhere later on we don't have to go back and add that default color to the styling. Thanks so much, as soon as that change is made I'll go ahead and merge this PR

leosoaivan:
@MxOliver I must have had my notifications turned off, because I never saw your last message. So sorry for the delay :pray:

Issue #342, opened 2019-10-20
Setup 'bundler-audit' gem


rgenchev:
I think that it's a good idea to install `bundler-audit` gem (https://github.com/rubysec/bundler-audit).

@MxOliver ?

kalimar:
@Rgenchev can you share some notes on what the gem is meant to do? Thanks

rgenchev:
@kalimar, sure :)

A `bundle audit check --update` command could be added to the CI and when building the application it will check for vulnerable gem versions in the `Gemfile.lock` file.

Let's say that there is a vulnerability in a specific gem but we don't know that. The build will fail and will suggest updating the corresponding gem to a specific safe version where the vulnerability is fixed.

Here is a detailed description: https://github.com/rubysec/bundler-audit#features
And some examples: https://github.com/rubysec/bundler-audit#synopsis

rgenchev:
I created https://github.com/Terrastories/terrastories/pull/363 that should close this issue.

I used `bundler-audit` and found vulnerabilities in 2 gems (`loofah` and `rubyzip`). In the PR I updated them.

PR #363, opened 2019-10-26
[#342] Setup 'bundler-audit' gem and fix vulnerabilities

rgenchev:
Closes #342 

In this PR I set up `bundler-audit` gem. I found 2 vulnerable gem versions and updated them.

PR #387, opened 2020-01-07
Revert "Revert "[#342] Setup 'bundler-audit' gem and fix vulnerabilities""

lauramosher:
Reverts Terrastories/terrastories#386 && Reopens original PR #342 

It looks like the build on Heroku succeeded so I will test this change locally and see if I can reproduce the issue that @MxOliver was having.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,7 @@ group :development, :test do
   gem 'pry-rails'
   gem 'guard-rspec'
   gem 'shoulda-matchers'
+  gem 'bundler-audit'
 end
 
 group :development do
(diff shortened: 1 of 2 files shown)


lauramosher:
This is a revert undoing the change in #386 and redoing the changes in #363.

I've verified that the changes were successfully deploying to heroku && work locally after running a new build to persist the new gem to the containers on the local environment. See the document linked on common errors in https://github.com/Terrastories/terrastories/blob/master/SETUP.md#having-troubles-check-our-common-errors--gotchas

PR #397, opened 2020-03-19
Allowing online image and video uploads to S3

mirandawang:
This PR includes the configuration necessary to allow the app to use ActiveStorage to upload images and videos (and other media files) to the S3 bucket we have set up. 

Its primary use is for the Heroku demo app, since Heroku does not support a local filesystem.

Generally, we expect development and in-field environments to use the local filesystem for ActiveStorage. You need the encrypted key in order to set the amazon configuration, and I've also included a comment saying not to switch it unless you're sure. So hopefully that dissuades most. :)

I've only included looking up the ENV variable in the production config file anyways, so in theory, you shouldn't be able to use the S3 bucket in development.

Current issue: 
In order for the images to show up properly in the admin dashboard, I needed to install imagemagick on the container itself. Is there a way I can get it to stay there permanently?
I needed to run `docker-compose exec web /bin/bash ` then `apt-get update && apt-get install imagemagick` for it to work.

This branch is currently deployed to Heroku if you'd like to see it in action without pulling!
I didn't have to install imagemagick on Heroku -- it just worked there.

To test these changes:
Pull this branch down.

Ask me to send you the secret keys for S3 config!

Then you'll need to create your Rails credentials (https://www.viget.com/articles/storing-secret-credentials-in-rails-5-2-and-up/)

After that, run these commands to install imagemagick on the container:
```
docker-compose exec web /bin/bash
apt-get update && apt-get install imagemagick
```

If you have media in your database currently, things will break if you change your ENV variable to amazon immediately. You will need to clear out all the rows first!
You can do that by running the following commands:
```
docker-compose exec web /bin/bash 
rails console
Story.delete_all
Speaker.delete_all
Place.delete_all
SpeakerStory.delete_all
```

Next, copy the line in `config/production.rb` and place it in `config/development.rb` in the same spot for active storage.

Change your ENV variable for `ACTIVE_STORAGE_SRC` to `amazon` . Restart the rails container.

You can create some rows in the admin dashboard and upload images and videos! Things should show up in the map and sidebar, and the files should show up in the S3 bucket :sparkles:

Follow ups:
* Rake task that can reset your DB for you (delete all rows)
* Validation for the ENV variable is set to a value that exists in storage.yml


.gitignore
```diff
@@ -64,3 +64,6 @@ yarn-debug.log*
 /node_modules
 yarn-debug.log*
 .yarn-integrity
+
+# Ignore master key for decrypting credentials and more.
+/config/master.key
```
lauramosher, line 69: This is also on line 19 of this file, do we need to add it again?

config/boot.rb
```diff
@@ -1,4 +1,4 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
-require 'bootsnap/setup' # Speed up boot time by caching expensive operations.
+# require 'bootsnap/setup' # Speed up boot time by caching expensive operations.
```
lauramosher, line 4: I noticed we commented out the bootsnap gem and the setup here — is that required for the S3 stuff?

robbkidd: I am 👍 on omitting bootsnap for the time being. bootsnap is a caching mechanism to speed up start times. As a cache, it introduces all of the weirdnesses of caching when it is not clear that Terrastories has a problem bootsnap caching solves.

config/storage.yml
```diff
@@ -7,12 +7,12 @@ local:
...
-#   access_key_id: <%= Rails.application.credentials.dig(:aws, :access_key_id) %>
-#   secret_access_key: <%= Rails.application.credentials.dig(:aws, :secret_access_key) %>
-#   region: us-east-1
-#   bucket: your_own_bucket
+amazon:
+  service: S3
+  access_key_id: <%= Rails.application.credentials.dig(:aws, :access_key_id) %>
+  secret_access_key: <%= Rails.application.credentials.dig(:aws, :secret_access_key) %>
+  region: "us-east-2"
+  bucket: "terrastories-demo"
```
robbkidd, line 15: Region and bucket should maybe be configurable through `ENV.fetch("ENV_VAR", "a_sensible_default")`. But they _aren't_ secrets, so they don't need to go through encrypted credentials.

config/boot.rb
lauramosher: Works for me! One fewer dependencies

.gitignore
mirandawang: Whoops missed that!

Issue #407, opened 2020-04-03
[Bug] [CMS] Photos for Speakers and Places are not rendering in the Administrate back end


rudokemper:
When adding photos for Speakers and Places via either the importer or by uploading them directly, these are working fine on the sidebar and map marker popups, but the image tags in the Administrate back end are not working. This seems to be a file path issue on the Administrate side.

Let's fix this so that the images are rendering correctly in Administrate.

PR #425, opened 2020-04-09
FIX: Render photos for Speakers and Places in the Administrate back end

GQuirino:
This PR fixes https://github.com/Terrastories/terrastories/issues/407.
BUG: Administrate gem was not able to render images from `active_storage`
FIX: Add gem `mini_magick`, and install dependencies into docker container

### DIFF
--- a/rails/Dockerfile
+++ b/rails/Dockerfile
@@ -14,6 +14,7 @@ RUN apt-get -y update && \
   git-all \
   curl \
   ssh \
+  imagemagick \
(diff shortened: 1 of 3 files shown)


FeminismIsAwesome:
@GQuirino this is looking good, just needs to be updated because we recently moved the rails dockerfile into a Rails directory in preparation for re-introducing offline mode back to the master branch.

Cheers,
Ian

GQuirino:
done :+1:

PR #432, opened 2020-04-25
Minimum viable e2e test framework

bhaibel:
- run a separate test-only Rails server
- e2e tests live in their own docker container
- use Cucumber w/ page objects to focus e2e tests on user
  goals rather than developer implementation details

Co-authored-by: slaloggia sarah.laloggia@gmail.com


FeminismIsAwesome:
LGTM

docker-compose.yml
```diff
@@ -33,6 +33,49 @@ services:
...
+
+  selenium:
+    # Debug version enables VNC ability
+    image: selenium/standalone-chrome-debug:3.0.1-germanium
+    # Bind selenium port & VNC port
+    ports: ['4444:4444', '5900:5900']
+    logging:
+      # Disable noisy logs.
+      driver: none
+    # bump this number up if pages are crashing
```
FeminismIsAwesome, line 62: haha normally dont like comments, but here they feel totes appropriate

FeminismIsAwesome:
lgtm post changes too. A question more for learning than about approval of PR:
with the e2e being a seperate test server, how does one set up data? like if they're going to see a list of places. Is the idea that youd only generate that data via user steps, or would there be a way to hook into the factories still?

Issue #469, opened 2020-07-13
Minor adjustment to image rendering in the marker pop ups


rudokemper:
Let's make the photo fill the full width of the pop up, instead of floating to the left of the text fields (as in the below screenshot). The text can go underneath the image. Let's also ensure that image proportions are respected, not shrunk as below.

![image](https://user-images.githubusercontent.com/31662219/87341238-4887cb00-c517-11ea-8412-724b22932ed3.png)

Current pop up photo rendering can be seen on the live demo: http://terrastories.herokuapp.com/en/home

Related follow up issue: https://github.com/Terrastories/terrastories/issues/350

ohamuy:
I can do this!

ohamuy:
I have already solved this problem if I can claim it!

rudokemper:
@ohamuy assigned!

PR #471, opened 2020-07-14
Pop-up image update

ohamuy:
This fixes issue #469. These are the pop-ups before: 
Screen Shot 2020-07-13 at 1 23 15 PM
Screen Shot 2020-07-13 at 1 23 02 PM

After Changes:
Screen Shot 2020-07-13 at 1 22 25 PM
Screen Shot 2020-07-13 at 1 22 19 PM

All images are in their original proportions, the container expands with the picture.

### DIFF
--- a/rails/app/assets/stylesheets/components/_balloon.scss
+++ b/rails/app/assets/stylesheets/components/_balloon.scss
@@ -1,19 +1,17 @@
 .ts-markerPopup {
   &-content {
-    display: flex;
     justify-content: center;
     padding: 0.5rem;
     color: $default-text;
 
     img {
-      width: 100px;
-      height: 100px;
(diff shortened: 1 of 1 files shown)


rails/app/assets/stylesheets/components/_balloon.scss
```diff
@@ -1,19 +1,17 @@
 .ts-markerPopup {
   &-content {
-    display: flex;
     justify-content: center;
```
lauramosher, line 3: Since we are removing `display: flex` we should also remove this — this CSS attribute won't do anything without display being set to flex!

ohamuy: It has been updated! All good to go

Issue #461, opened 2020-07-05
Fix rspec tests


lauramosher:
Right now, if you run `docker-compose run web-test rspec`, there are a number of failing specs.

```
Finished in 14.18 seconds (files took 3.24 seconds to load)
84 examples, 21 failures
```

This issue is to work through those failing specs and make them pass.

slaloggia:
I'll take this! A lot of them appear to be due to a fix I implemented for a different issue anyway, so I should have updated the specs then.

slaloggia:
The .import_csv tests were failing bc the Place model is validating lat & and long. Should those be allowed to be blank or do we only want Places to be created with a latitude and longitude?

lauramosher:
> Should those be allowed to be blank or do we only want Places to be created with a latitude and longitude?

I would say allowed to be blank but I am not sure what the original ticket was asking. @Terrastories/terrastories-stewards pinging to see if you have any input on the preference for lat/long requirements.

slaloggia:
Is there anything else that wasn't addressed by that PR, or can this be closed?

lauramosher:
@slaloggia This can be closed. I'm so used to them being closed automatically. 🙈 

In the future, you can tag a PR with one of the github keywords so it will automatically close this out. In the one linked above, instead of using "Address issue #461" you can use "fixed #461" or "closes #461." Here are all of the documented keywords that will close the associated issue: https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Thank you so much for your work on this!

PR #475, opened 2020-07-20
Fix rspec tests such that all tests run and pass

slaloggia:
- Addresses issue #461 
- Corrects syntax in several files so that there is only one expect statement per 'it' block
- Adds speakers to all necessary Story parameter objects
- Makes lat & long optional for Place creation
- Adds basic association tests for the Curriculum model

### DIFF
--- a/rails/spec/models/story_spec.rb
+++ b/rails/spec/models/story_spec.rb
@@ -5,8 +5,6 @@ RSpec.describe Story, type: :model do
     it { should have_many :speaker_stories }
     it { should have_many :speakers }
     it { should have_and_belong_to_many :places }
-
-    # Issue 406 requested that interviewer and interview location be made optional for story creation
     it { should belong_to(:interview_location).optional }
     it { should belong_to(:interviewer).optional }
   end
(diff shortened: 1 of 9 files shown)


rails/spec/models/story_spec.rb
```diff
@@ -14,45 +14,51 @@
...
-      expect(story.language).to eq csv[7]
-      expect(story.permission_level).to eq "anonymous"
-      expect(story.media.first.filename.to_s).to eq csv[8]
+    describe 'delegates all attributes to decorator' do
+      before(:all) do
+        @fixture_data = file_fixture('story_with_media.csv').read
+        described_class.import_csv(@fixture_data)
+      end
+      let!(:story) {described_class.last}
+      # puts story.title
```
FeminismIsAwesome, line 34: @slaloggia whats going on here? are these comments just need to be removed?

slaloggia: They are indeed. Will fix.

slaloggia: Should be good to go. Double-checked all the spec files for other stray comments and removed a couple in other places.

PR #478, opened 2020-07-28
add model and dashboard for themes, added controller logic to only al…

kev-kev:
…low for one active theme at a time

The attributes for themes are an "active" boolean, and a "background_img" that's passed as a string.

After signing into the admin page, you can create a custom theme by following these steps:
- Select "Themes" from the sidebar, then click "New Theme" 
- Under "Background Image", enter the URL for your image
- Tick the checkbox next to "Active" to make this theme active upon creation

When a theme is set to active while being created or edited, it causes any other active theme to be made inactive. This makes it so only one theme can be set to active at any time. 

After setting an active theme, if you view the welcome page, you will be able to see your active theme's background image as the background for the welcome page.


rails/app/controllers/admin/themes_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    def update
+      if theme_params[:active]
+        deactivate_prev_theme
+      end
+      super
+    end
+    
+    def deactivate_prev_theme
+      prev = Theme.find_by(active: true)
+      prev.update(active: false)
```
lauramosher, line 20: I think if we change these lines to this one:

```suggestion
      Theme.where(active: true).update_all(active: false)
```

It would stop any errors from trying to run `update` when no record is active found. (`update_all` will still run on an empty relation and commit no updates). It would also make it so if we somehow ended up with multiple themes active, it would ensure ALL of them are marked as inactive.

If we would rather leave it the same, we can do a guard check against no-previously active theme on line 20 to avoid errors:

```suggestion
      prev = Theme.find_by(active: true)
      prev.update(active: false) if prev.present?
```

rails/db/migrate/20200726232543_create_themes.rb
```diff
@@ -0,0 +1,10 @@
+class CreateThemes < ActiveRecord::Migration[5.2]
+  def change
+    create_table :themes do |t|
+      t.string :background_img
+      t.boolean :active
```
lauramosher, line 5: Lets change this to set a database constraint to `null: false` - this way when we are utilizing the boolean in Rails, it'll always be a true boolean (instead of a 3-way "boolean" - true / false / null [falsey]).

```suggestion
      t.boolean :active, null: false, default: false
```

Here is more thoughts on the threestate boolean problem: https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem.

rails/app/views/welcome/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="welcome">
+<div class="welcome" style="background-image: url(<%= @theme ? @theme.background_img : 'welcome-bg.jpg' %>) " > <%#no-repeat center center fixed%>
```
lauramosher, line 1: I noticed that when there is no theme set, the default welcome image is missing / a blank white background.

I believe that's because now that we are referencing the image from the root (rather than relative assets in the stylesheet), we need to specify the assets folder.

```suggestion
<div class="welcome" style="background-image: url(<%= @theme ? @theme.background_img : 'assets/welcome-bg.jpg' %>) " >
```

rails/app/controllers/admin/themes_controller.rb
```diff
@@ -0,0 +1,27 @@
+require 'pry'
+module Admin
+  class ThemesController < Admin::ApplicationController
+    def create
+      if theme_params[:active]
```
lauramosher, line 5: So, this check and the one on line 12 will *always* run the `deactivate_prev_theme` code even if active was not checked (is set to false) in the form.

This is because the form input is sending back "0" / "1" for false / true if the checkbox was checked or not, and so an object check will always return true because it's a string.

The quickest way to resolve this is to check if `theme_params[:active] == "1"`, which is the equivalent to true.

kev-kev:
Hey @lauramosher - thanks for the review! I made the changes you suggested and pushed them up. Let me know if everything looks good now!

lauramosher:
Yes! The click through I did works great.

There are a couple of unrelated things that need to get resolved before a merge can happen:

- It looks like the `.env.example` accidentally got removed in this change. Can you please add that back in?
- It also looks like there are conflicts in the `Gemfile.lock` that need to be resolved before we can merge. A refresh with the main branch should resolve them. I think you can also remove the Gemfile.lock changes from your branch entirely since you didn't add any Gems yourself.

Let me (or one of the Stewards) know if you need help navigating this.

Thank you again for the contribution!

kev-kev:
@lauramosher 
Ah gotcha- I just copied over the .env.example and gemfile.lock files from the most recent version of the app, so I think it should be good now.

rails/app/assets/stylesheets/components/welcome.scss
```diff
@@ -1,5 +1,6 @@
 .welcome {
-  background: image-url('welcome-bg.jpg') no-repeat center center fixed;
+  background: no-repeat center center fixed;
+  // background: image-url('welcome-bg.jpg') no-repeat center center fixed;
```
FeminismIsAwesome, line 3: @kev-kev can you remove this comment?

kev-kev:
@FeminismIsAwesome Thanks for catching! It's now removed.

PR #481, opened 2020-08-01
Nested has many test clean

slaloggia:
Attempts to reimplement the nested table of stories in the admin Show page for Curriculum.


slaloggia:
Example show page with nested Stories table. Additional attributes and label changes for the table are easy to accomplish, so just let me know if you'd like tweaks!

Screen Shot 2020-08-03 at 3 05 23 PM

FeminismIsAwesome:
@slaloggia the code renders fine, but got some conflicts to resolve

slaloggia:
@FeminismIsAwesome, those current conflicts are necessary changes for the current code to work. Part of the reason the nested forms weren't working was because that gem was relying on the current version of Administrate, and we were using an older version. And then the navigation partial had to be updated to work with that version as well. Let me know if you want to discuss further or if there's anything else I can do to resolve on my end.

rails/app/assets/stylesheets/admin/_main_content_style.scss
```diff
@@ -35,7 +35,7 @@
 .form {
     margin-left: 2rem;
     border: 4px  #0f566b solid;
-    width: 300px;
+    width: auto;
```
FeminismIsAwesome, line 38: does this leave a consequence on other admin forms? thats a bit of a sweeping css change

rails/app/views/admin/curriculums/_form.html.erb
```diff
@@ -0,0 +1,48 @@
...
+      <ul>
+        <% page.resource.errors.full_messages.each do |message| %>
+          <li class="flash-error"><%= message %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
+
+  <% page.attributes.each do |attribute| -%>
+    <% unless excluded_resources.include?(attribute.name) %>
```
FeminismIsAwesome, line 32: rather than edit all renders for all time, is there a way we could rewrite just the curriculum related pages?

rails/app/views/admin/curriculums/edit.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    <%= link_to(
+      t("administrate.actions.show_resource", name: page.page_title),
+      [namespace, page.resource],
+      class: "button",
+    ) if valid_action?(:show) && show_action?(:show, page.resource) %>
+  </div>
+</header>
+
+<section class="main-content__body">
+  <%= render "form", page: page, excluded_resources: excluded_resources %>
```
FeminismIsAwesome, line 30: i think rather than hijacking the shared form, building a custom version of the form specific to curriculum stories might be better

rails/app/views/fields/text/_index.html.erb
```diff
@@ -0,0 +1,13 @@
+<%#
+# Text Index Partial
+This partial renders a text attribute
+to be displayed on a resource's index page.
+By default, the attribute is rendered as a truncated string.
+## Local variables:
+- `field`:
+  An instance of [Administrate::Field::Text][1].
+  A wrapper around the Text pulled from the database.
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Text
```
FeminismIsAwesome, line 10: remove this file

FeminismIsAwesome (approved):
Looks good to me! Merging.

Issue #476, opened 2020-07-21
[Import/export] CSV importer fails for rows without a media attachment


rudokemper:
The `CSV importer` seems to fail whenever it comes across a row that does not have a media attachment. See screenshot below.

This issue came up when importing `Places` using the importer. All of the Places imported without a problem until the importer came across a row for a Place without a media attachment. Then I got the following message.

If there is no media attachment for a Place (or other object), the CSV importer should just move on to the next row, without attaching anything.

![image](https://user-images.githubusercontent.com/31662219/88093889-a85a2380-cb60-11ea-8b08-4e377a141c37.png)

rudokemper:
This is what the CSV file I'm using looks like. The importer fails and gives the above message when it comes across the last row in this screenshot, `Afitimama`

![image](https://user-images.githubusercontent.com/31662219/88094372-72696f00-cb61-11ea-9f04-04d2a46d67be.png)

rudokemper:
I can confirm that the same issue arises for Stories without a media attachment.

![image](https://user-images.githubusercontent.com/31662219/88095329-ebb59180-cb62-11ea-9a49-927ebf811425.png)

PR #482, opened 2020-08-04
fix for issue #476

kathenry:
Allows csv files to be imported without media attached. If fix is correct, can easily be applied to stories as well.

Co-Authored-By: Lucas Alcantara <lukmin1999@users.noreply.github.com>
Co-Authored-By: Oliver Hamuy <o.hamuy@wustl.edu>

### DIFF
--- a/rails/app/decorators/place_photo_decorator.rb
+++ b/rails/app/decorators/place_photo_decorator.rb
@@ -4,7 +4,7 @@ class PlacePhotoDecorator
   end
 
   def attachable?
-    @filename != "" && File.exist?(path)
+    @filename.present? && File.exist?(path)
   end
 
   def blob_data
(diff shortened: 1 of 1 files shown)


rails/app/decorators/place_photo_decorator.rb
```diff
@@ -4,7 +4,7 @@ def initialize(filename)
   end
 
   def attachable?
-    File.exist?(path)
+    @filename != "" && File.exist?(path)
```
FeminismIsAwesome, line 7: in ruby, there's a preference for a method that catches all the empty-variations
@filename.present? 
https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html

FeminismIsAwesome:
```
@filename.present? && File.exist?(path)
```

PR #483, opened 2020-08-04
Update docker and docker-compose config and scripts

lauramosher:
## Summary of Changes

- Updates `docker-compose.yml` to use more recent configuration for dependent builds/images.
- Removes `e2e` test script in favor of root test script that runs the command via run.
- Adds global test script that runs e2e specs and rspec unit tests
- Renames and updates `script/test` to `script/test.sh` for Windows running capability. Windows PowerShell requires a file extension to run.
- Removes `rails/scripts/wait-for-it.sh` as part of new `docker-compose` setup

### File Extension Naming (why is there `.sh` sometimes, but not all of the time?)

I've made a change to how we name / utilize file extensions in the app:

1. For any script that is run via docker, the file extension has been _removed_. *nix does not require file extensions to run bash/shell scripts, and so any script that we'd run in docker and only docker, I've removed the file extension.
1. For any script that is run via command-line outside of docker (<root>/script/*) have a `.sh` file extension. This is so when run in Windows Powershell, it will be able to be executed. Windows will error if it doesn't know / recognize the file extension.

This is more for a cognitive separation: if the script doesn't have a .sh extension, there is a good chance you shouldn't be running it.

### First time startup

I made a few superficial changes to boot up.

Now, when the server is first started, there is a message before boot up that we are starting the server.

There is a second message that displays after initialization and before it's fully finished booting. This flushes standard out for Docker and shows progress for when the server is actually bootable.

```
➜  terrastories git:(lm-docker-updates) docker-compose up web
Starting terrastories_db_1 ... done
Starting terrastories_web_1 ... done
Attaching to terrastories_web_1
web_1           | Starting Rails Server...
web_1           | => Booting Puma
web_1           | => Rails 5.2.3 application starting in development
web_1           | => Run `rails server -h` for more startup options
web_1           | Initialized
web_1           | Puma starting in single mode...
web_1           | * Version 3.12.6 (ruby 2.5.1-p57), codename: Llamas in Pajamas
web_1           | * Min threads: 5, max threads: 5
web_1           | * Environment: development
web_1           | * Listening on tcp://0.0.0.0:3000
web_1           | Use Ctrl-C to stop
```

### Commands/Startup

At this point, *all* of the expected docker commands should continue to function as expected, such as build/up. Please run through setup instructions again and ensure that things still work as expected.


docker-compose.yml
```diff
@@ -17,39 +17,34 @@ services:
...
-    links:
-      - db
-    ports:
-      - 3000:3000
-    command: ["./scripts/wait-for-it.sh", "db:5432", "--", "./scripts/start_rails.sh"]
     volumes:
       - ./rails:/opt/terrastories:cached
 
+  web:
+    <<: *x-web-common
```
robbkidd, line 32: 👍  Aww yeah, YAML merging!

PR #488, opened 2020-08-11
♻️ Replacing react-virtualized with react-list in StoryList ♻️ (Fixes #228)

### LINKED ISSUE #228: [UI] Card height not properly adjusting to content
To replicate the issue:

1. Add two videos to a single story
2. Use filters so that that story appears as the first one in the sidebar
3. Change filters
4. The new first card is now the same height as the previous card, despite not having two videos, so the effect is that there is too much whitespace.

We should make sure that the Card height is determined by the size of the media it contains.

mirandawang:
This PR replaces `react-virtualized` entirely with `react-list` 
& adds a small refactor to move Story into its own component

This should hopefully make working w/ the StoryList MUCH more simple and intuitive.

*What's the diff between the two libraries?*
`react-virtualized` creates virtualization whereas `react-list` is just an infinite scrolling list. Virtualization optimizes for performance by replacing the currently rendered divs with new content, rather than actually scrolling and adding new DOM nodes to the top. As you scroll up and down, the Virtualized list calculates which rows are in view, and only renders those. 

This works best with more static data, however, and since the heights of each of our stories can differ, it may not be the best suited library for our use. We also want to have the ability to filter and change the stories easily, which means that the list rows at specific spots need to be able to adjust their heights accordingly.

With `react-list`, this works better because everything exists in the DOM and we let the browser do its calculations as normal. But we lose the benefit of the virtualization performance optimization.

However! I tested this out with about ~300 rows & a good amount of videos. It didn't appear to affect the performance of the app too much -- I think our bottleneck is currently the map tiles still. So I think this is a worthwhile change to make!

✨ Pros ✨
* Less bloat in the app since we don't need all the features in react-virtualized anyway
* Easier to understand the interface with react-list because there's less going on
* Still can have dynamic row heights + infinite scrolling at relatively the same performance
* WAY less bugs and weird rendering issues!
* Happier developer because things don't break randomly

Fixes: 
Bug w/ card heights rendering incorrectly after filter
Bug w/ images in stories

In the future, if we notice that the list is becoming a performance issue, then we can revisit using a virtualized list. The same author has more recently come out with a library called `react-window` which should be a smaller version of the previous. We could keep an eye on this!

**To test these changes:**
Pull my branch
Run `docker-compose build` to get new package changes
Add some more videos and images to the Stories
Mess around w/ filters
No weird rendering issues!

### DIFF
--- a/rails/app/javascript/components/App.jsx
+++ b/rails/app/javascript/components/App.jsx
@@ -93,7 +93,6 @@ class App extends Component {
         }
       }
     });
-    console.log(filterMap);
     return filterMap;
   };
 
@@ -178,7 +177,7 @@ class App extends Component {
   handleFilterItemChange = option => {
     if (option === null) {
       this.resetStoriesAndMap();
-    } else {
+    } else if (this.state.filterCategory !== null) {
       const item = option.value;
       this.handleFilter(this.state.filterCategory, item);
       this.setState({ filterItem: item });
@@ -224,7 +223,6 @@ class App extends Component {
   };
 
   handleMapPointClick = (point, stories) => {
-    console.log(point);
     this.showMapPointStories(stories);
     const framedView = { center: point.geometry.coordinates };
     this.setState({ activePoint: point, framedView });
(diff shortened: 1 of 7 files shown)


rails/app/javascript/components/App.jsx
```diff
@@ -93,7 +93,6 @@ class App extends Component {
         }
       }
     });
-    console.log(filterMap);
```
FeminismIsAwesome, line 96: yeeeeee

lauramosher (approved):
🎉

PR #497, opened 2020-09-03
add jest specs with readme directions

FeminismIsAwesome:


mirandawang:
Code LGTM, but I get this error when I try to run the command you put in the README:
`OCI runtime exec failed: exec failed: container_linux.go:349: starting container process caused "exec: \"jest\": executable file not found in $PATH": unknown`

I tried rebuilding the container and run the web setup script. But still same error.

FeminismIsAwesome:
@mirandawang I made a mistake when I was adding the readme. great catch
`docker-compose exec web yarn jest` should be the command instead

mirandawang (approved):
I was able to run the command, tests are passing & it LGTM! :D

documentation/DEVELOPMENT.md
```diff
@@ -41,6 +41,12 @@ You can run RSpec and e2e tests with
 script/test.sh
 ```
 
+We also support Javascript unit testing, with Enzyme for snapshots. 
+
+```
+    docker-compose exec web yarn test
```
lauramosher, line 47: The `script/test.sh` runs both the Rails unit tests and the e2e integration tests. It might be nice to _also_ run this in that test script so it's full gamut, and then detail how they can run each individual suite separately (like this one).

lauramosher (approved):
LGTM :shipit:

PR #519, opened 2020-09-24
514 - Add jest/enzyme tests for Card.jsx

leosoaivan:
## Description
This PR adds basic snapshot testing and high-level unit testing for the `Card.jsx` component. I mainly focused on side-effects and unit interactions to avoid white-box testing. I would otherwise recommend Cypress for fully testing integrations between Card, Filter, and StoryList, and testing viewport size behaviors.

Open to suggestions or re-working to go deeper, though!


FeminismIsAwesome (approved):
I think this is a great start!

PR #527, opened 2020-10-03
#510 - Add Sort.jsx specs

jaimevelaz:
## Description
This PR adds specs to `Sort` component, using snapshot testing and _naive_ unit testing for the sorting algorithm executed when the Select in changed.

Triggering changes over `Select` component is a little bit cumbersome but after some research I couldn't found a better way of doing it using `enzyme`, I'll be happy to update and improve anything from this PR.

Closes https://github.com/Terrastories/terrastories/issues/510

### DIFF
--- a/rails/spec/javascript/components/Sort.spec.js
+++ b/rails/spec/javascript/components/Sort.spec.js
@@ -1,0 +1,67 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import Sort from '../../../app/javascript/components/Sort';
+
+describe('Sort component', () => {
+  beforeEach(() => {
+    global.I18n = {
+      t: jest.fn(),
+    };
+  });
+
+  const corinneStory = {
+    created_at: '2020-10-03T11:13:23.037Z',
+    title: "Corinne's testimonial",
+  };
+
+  const rudoStory = {
+    created_at: '2020-10-01T11:13:23.013Z',
+    title: "Rudo's testimonial",
+  };
+
+  it('Displays correctly', () => {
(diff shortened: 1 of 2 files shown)


rails/spec/javascript/components/Sort.spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+  it('Displays correctly', () => {
+    const wrapper = shallow(<Sort stories={[corinneStory, rudoStory]} />);
+
+    expect(wrapper).toMatchSnapshot();
+
+    expect(global.I18n.t).toHaveBeenCalledWith('sort_stories');
+    expect(global.I18n.t).toHaveBeenCalledWith('most_recent');
+    expect(global.I18n.t).toHaveBeenCalledWith('alphabetical');
+    expect(global.I18n.t).toHaveBeenCalledWith('reversed_alphabetical');
```
mirandawang, line 30: What do you think of moving out these hardcoded strings into a const array? Then we can easily update the text or add to it if the Sort menu changes. You could define an array of sortValues which have all these strings in them and then use a for loop to call the expect statement on each.

mirandawang (commented):
Thanks for these tests! I pulled down your branch and everything is passing for me. I just had 1 suggestion about the display test of the various sort value strings. :)

jaimevelaz:
> Thanks for these tests! I pulled down your branch and everything is passing for me. I just had 1 suggestion about the display test of the various sort value strings. :)

Thanks for the review @mirandawang, I liked your suggestion and I've updated the code with it.

mirandawang:
Awesome! Thanks so much @jaimevelaz !

mirandawang (approved):

Issue #505, opened 2020-09-03
[Rails][Administrate] Add ability to upload/include a custom logo in Themes


mirandawang:
Since the terrastories app can be used by more than just one community, we would like it to be flexible and customizable through our concept of Themes.

There is a Theme model which exists and currently can change visual elements of the app, as well as some logic to make sure only one Theme is active at a time.

For this issue, let’s extend the model so you can also upload a custom logo. This should show up under the box on the Welcome page here:
![Screen Shot 2020-09-03 at 5 35 36 PM](https://user-images.githubusercontent.com/8590508/92176158-e4dd9800-ee0b-11ea-897f-cb41de41b4f6.png)

You can find the code here: https://github.com/Terrastories/terrastories/blob/master/rails/app/views/welcome/index.html.erb#L30

Users should be able to add a custom logo to a Theme in the admin panel via uploading through ActiveStorage. You can look at the Place or Story model as an example as to how we use ActiveStorage to attach images.

To login to the admin panel, you can use the default terrastories admin account which is set up in the seeds.rb file.

We should also add some validations to make sure that this field can only take image format uploads.

Acceptance Criteria:
Theme can be created/edited with a logo field in the admin panel
The active theme’s logo appears on the Welcome page underneath the box.
If a user uploads a file that is not an image, show an error on the admin page.

Bonus points: Add unit tests

tauhir:
Can I pick this up? I've submitted a PR for #506 and this is quite similar.

mirandawang:
Hi @tauhir that would be great! And will review your PR ASAP :)

PR #545, opened 2020-10-12
506: sponsor logos in theme

### LINKED ISSUE #506: [Rails][Administrate] Add ability to upload sponsor logos in Theme
We have a Theme model which allows us to change different UI features of the Terrastories app for different communities that use it.

Admin are able to update the Theme properties through the admin panel and have one active Theme at a time.

To login to the admin panel, you can use the default terrastories admin account which is set up in the seeds.rb file.

Let’s add the ability to set sponsor logos which will change add images on the welcome screen : 
Screen Shot 2020-09-03 at 5 35 20 PM

The sponsor images should go underneath the box and below the container for the custom logo here https://github.com/Terrastories/terrastories/blob/master/rails/app/views/welcome/index.html.erb#L30 in it's own div. Let's add an extra span with the text "Built with :heart: by" above the sponsor images. Ideally this is centered but feel free to design what looks nice to you!

The Ruby For Good logo should always be included by default! If there are no sponsor logos attached to the Theme, then only the R4G logo should be displayed. Here's the R4G logo we can use:
rubyforgood

If the user attempts to upload a non image format file, we should show an error.

Acceptance Criteria:
Theme can be created/edited with multiple sponsor logos attached
The sponsor logos appear,along with the R4G logo, on the initial Welcome page
If there is no theme active or no sponsor logos attached, only the R4G logo is shown
There's some text that says "Built with :heart: by" above the sponsor logos.
Errors are shown on the admin page if non images are uploaded.

Bonus points: Add unit tests

tauhir:
This PR resolves #506 
This adds the ability to add sponsor logos to a theme. I added the activestorage-validator gem which makes model validation easier. If its not wanted, I can remove the gem and change my approach to match rails/app/models/place.rb


rails/spec/models/theme_spec.rb
```diff
@@ -2,4 +2,11 @@
 
 RSpec.describe Theme, type: :model do
   pending "add some examples to (or delete) #{__FILE__}"
+
+  let(:theme) {Theme.new()}
+  it "can add sponsor logo" do
+    theme.logos.attach(io: File.open("./spec/fixtures/media/terrastories.png"), filename: 'file.pdf')
+    theme.valid?
```
FeminismIsAwesome, line 9: not blocking but curious: why not 
expect(theme).to be_valid
?

FeminismIsAwesome (approved):
Looks good to me! Merging. Non blocking for-my-curiosity comment.

rails/spec/models/theme_spec.rb
tauhir: That makes a lot more sense than my method. I think I initially compared the message to the expected message and I ended up at that. When I picked this up, I wasn't aware of #505 which adds another logo, could I make a new issue to update the model to sponsor_logo or something similar? I'll update this test too.

FeminismIsAwesome: @tauhir that sounds like a good consideration! make it an issue and I'll bring it up with the rest of the team to see if that makes sense, makes sense to me though. Definitely feel free to make a follow up PR.

PR #549, opened 2020-10-15
[#547] Add model specs for Curriculum

### LINKED ISSUE #547: [Tests] Add model specs for Curriculum
I noticed that we could add more specs for `Curriculum` model.

rgenchev:
Fixes #547


FeminismIsAwesome:
@rgenchev not sure these kind of tests add much value, there are other open issues around specific specs needed if that's what you want to add.

they feel like they're testing rails y'know?

PR #550, opened 2020-10-15
[#548] Add model specs for CurriculumStory

### LINKED ISSUE #548: [Tests] Add model specs for CurriculumStory
I noticed that we could add more specs for `CurriculumStory` model.

rgenchev:
Fixes #548


FeminismIsAwesome:
@rgenchev not sure these kind of tests add much value, there are other open issues around specific specs needed if that's what you want to add. 

they feel like they're testing rails y'know?

Issue #539, opened 2020-10-09
[Rails][Administrate] Ensure that duplicate Places cannot be added using the importer 


mirandawang:
For Places, admin are able to upload a CSV to bulk import data. However duplicate items can get added. Let’s update the import code so that duplicate items cannot be added.

Match on all fields. If this EXACT record already exists, do not recreate it. Just move on to the next row.

Sample CSV file: https://drive.google.com/file/d/1Ir6TC5THEoQKk-98JZ5KpTjXd-R6yMmA/view

Importer code for [Places](https://github.com/Terrastories/terrastories/blob/master/rails/app/controllers/places_controller.rb#L65)

You can log into the admin interface using the sample admin account from the db seeds.

Acceptance Criteria:
When using the CSV importer, duplicate rows that have the same exact fields do not get added twice. 
Add unit tests for this behavior in each place_spec.rb

jaimevelaz:
I've created a Pull request (https://github.com/Terrastories/terrastories/pull/551) to try to ensure places cannot be duplicated through the CSV importer, after adding the spec found that this cannot happen but having the spec may be useful for the future.

lauramosher:
Thanks for digging into this, @jaimevelaz — since the PR you opened was merged, I'm going to close this issue.

PR #551, opened 2020-10-19
#539 Ensure that duplicate Places cannot be added using the importer

jaimevelaz:
_Fixes_ https://github.com/Terrastories/terrastories/issues/539

This PR adds a spec to ensure places cannot be duplicated through the `Place` importer, if I understood correctly the issue the csv importer was already ensuring a single Place with same attributes is created by using `place = Place.find_or_create_by(decorator.to_h)` in line https://github.com/Terrastories/terrastories/blob/master/rails/app/models/place.rb#L17

### DIFF
--- a/rails/spec/models/place_spec.rb
+++ b/rails/spec/models/place_spec.rb
@@ -12,7 +12,7 @@ RSpec.describe Place, type: :model do
     end
 
     describe "does not create duplicated places" do
-      it "creates a single place for same file twice" do
+      it "creates a single place when importing the same csv twice" do
         @fixture_data = file_fixture('place_with_media.csv').read
 
         expect {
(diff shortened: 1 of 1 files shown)


FeminismIsAwesome:
@jaimevelaz i'm seeing a test, but it doesn't seem like the contents of the spec match the assertion, and missing the passing behavior to go with?

FeminismIsAwesome (changes requested):
make the test red-green by adding the behavior of not creating 2 on duplicates instead of just asserting 1 on creation, and fix importer code

jaimevelaz:
Thanks for the review @FeminismIsAwesome, I'll take another look to it, initially I thought building a CSV with two identical lines and asserting a single place is created would be enough to validate the defined behaviour but I'm probably missing something.

jaimevelaz:
> make the test red-green by adding the behavior of not creating 2 on duplicates instead of just asserting 1 on creation, and fix importer code

I've changed the spec to make it easier to read and understand. Feedback is welcome @FeminismIsAwesome

FeminismIsAwesome (approved):
awesome! merging.

Issue #542, opened 2020-10-09
[Rails][Administrate] Make errors visible when the Story importer fails


mirandawang:
For Stories, admin are able to upload a CSV to bulk import data. When errors occur, the process fails and the errors are not visible.

Let’s add better error handling so that the importer continues to try to import each row, gathering the errors along the way. Instead of failing when it hits the first error, return the errors at the end of trying to import all of the rows, if there are any. Then display them when the page reloads.

Sample CSV file: https://drive.google.com/file/d/1Ir6TC5THEoQKk-98JZ5KpTjXd-R6yMmA/view

Importer code for [Stories](https://github.com/Terrastories/terrastories/blob/master/rails/app/controllers/stories_controller.rb#L36)

Acceptance Criteria:
When importing, if there are errors, the importer does not fail completely, it tries to upload what it can and save the errors that occur.
The errors that occur during import are displayed on the admin page.
Valid rows should be able to get uploaded even if invalid rows exist in the same CSV.
Add unit tests for this behavior in  story_spec.rb

tauhir:
Could I pick this up?

rudokemper:
Yes please! :)
-------- Original message --------From: Tauhir Edwards <notifications@github.com> Date: 10/31/20  07:20  (GMT-05:00) To: Terrastories/terrastories <terrastories@noreply.github.com> Cc: Subscribed <subscribed@noreply.github.com> Subject: Re: [Terrastories/terrastories] [Rails][Administrate] Make errors visible when the Story importer fails (#542) 
Could I pick this up?

PR #559, opened 2020-11-07
#540, #541, #542 csv importer error handling updates

### LINKED ISSUE #540: [Rails][Administrate] Make errors visible when the Place importer fails
For Places, admin are able to upload a CSV to bulk import data. When errors occur, the process fails and the errors are not visible.

Let’s add better error handling so that the importer continues to try to import each row, gathering the errors along the way. Instead of failing when it hits the first error, return the errors at the end of trying to import all of the rows, if there are any. Then display them when the page reloads.

Sample CSV file: https://drive.google.com/file/d/1Ir6TC5THEoQKk-98JZ5KpTjXd-R6yMmA/view

Importer code for [Places](https://github.com/Terrastories/terrastories/blob/master/rails/app/controllers/places_controller.rb#L65)

You can log into the admin interface using the sample admin account from the db seeds.

Acceptance Criteria:
When importing, if there are errors, the importer does not fail completely, it tries to upload what it can and save the errors that occur.
The errors that occur during import are displayed on the admin page.
Valid rows should be able to get uploaded even if invalid rows exist in the same CSV.
Add unit tests for this behavior in  place_spec.rb

tauhir:
This PR resolves #540 , #541 & #542 :

I started working on #542 and realized that a general solution would be beneficial for Places, Speakers and Stories - I used a helper to do this. 

I've added a custom partial for the Administrate gems'  flash messages so that we can handle the errors and show them line by line. Flash messages can be added as a string as well as a hash which will be printed out as shown below: 
image

The only issue I came across is that there are no validations for the speaker models and therefore a blank speaker can be imported. I've still added the code and the tests but I've left them pending for now. I tested that it works by adding a `:validates speaker_name` to rails/app/models/speaker.rb

Hopefully its fine that I've done all three issues, if not, please let me know 😄


mirandawang (approved):

FeminismIsAwesome:
@tauhir looks like we have some merge conflicts! we can accept once those are fixed. sorry for the delay in review1

tauhir:
I had to update the admin places, speaker and story controllers with the csv logic I added in this branch because the base controllers were deleted in #560.

PR #560, opened 2020-11-09
Remove unused base controllers.

lauramosher:
With the exception of the import_csv routes, these controllers were
completely unused. This moves the import_csv code into their respective
admin controllers (and correctly auth walled to admins now!) and removes
all of the excess controllers, controller specs, request specs, views,
and routes.


mirandawang (approved):
:tada: Yay for all the deletes!

I tested your branch out locally, tried the import for Speaker, Place, and Story. Works fine. Clicked around in the app as well and everything seems to be working--  LGTM!

PR #561, opened 2020-11-12
add feature flipper and ability to hide modules that are experimental

FeminismIsAwesome:


rudokemper (approved):
Testing it out by adding `themes` to the array of excluded resources in line 23 of `rails/app/controllers/admin/application_controller.rb`. Themes successfully excluded from the admin menu thereafter.

rudokemper:
Closing this as we ended up handling feature flipping in a different way, but I think this was helpful as a reference! Thanks @FeminismIsAwesome and sorry for the delay; hope you're doing well!

PR #563, opened 2020-11-20
[348] Refactoring Map component + using Mapbox markers

mirandawang:
This PR refactors the Map component to start using Mapbox markers rather than HTML markers.

This should help the map load more performant.
Instead of adding a new DOM element for every marker, we add all the points as a data source and a layer on top of the Mapbox map.

All other changes are needed to re-bind existing events on to the new Mapbox markers. For Popups, I created a new JSX component that encapsulates its markup. There are a few changes here to how popups close and open, but I'm now keeping track of what active popup is open in the component state. Then I use that information for displaying or closing popups, since we only display 1 active popup at a time. 

Since we programatically close pop ups now when we open a new one, I created a new custom click handler for a custom close button on the popups. (There was no way to attach directly to just the close button of the popup, in React without refs. This seemed the best/cleanest solution). When closing a popup through the close button (as opposed to being closed since you clicked on a new one), we'll clear out all the filtered stories.

To test these changes:
1. Pull my branch down
2. Visit the Story map
3. Click around on Points, Stories, etc. everything should be working the same as it was before!


rails/app/javascript/components/Map.jsx
```diff
@@ -87,65 +93,54 @@ export default class Map extends Component {
...
-            marker.properties.type_of_place
-          }</div>
-            </div>
-          </div>`;
-        } else {
-          return `<h1>${marker.properties.name}</h1>
-          <div class="ts-markerPopup-content">
-            <img src=${marker.properties.photo_url} />
-            <div>
-              <div>${I18n.t("region")}: ${marker.properties.region}</div>
```
mirandawang, line 113: I'm a lil confused on how I18n.t was working in the JSX view previously. I removed it since it seemed it wasn't doing anything but I can put it back?

FeminismIsAwesome:
@mirandawang i thiiink this would work for offline mode, but did you confirm this?

would be my only concern with a mapbox marker

rudokemper:
I can now confirm this works perfectly with offline mode.

rails/app/javascript/components/Map.jsx
lauramosher: It looks like it does do SOMETHING. I wonder if there is a hookin for I18n library in the JavaScript / asset pipeline.

image

lauramosher (approved):
I'm going to merge this without the I18n change, we can come back through and readd that @mirandawang

PR #566, opened 2020-11-23
Update SETUP-OFFLINE.md with missing instructions

rudokemper:
We were missing one step in the SETUP-OFFLINE.md readme for running offline mode.


lauramosher (approved):
Seems legit :)

PR #572, opened 2020-12-12
Terrastories Multi-Instance Support

lauramosher:
This introduces Terrastories Multi-Instance support for online hosted communities.

It's big, but it needs to be because it re-wires just about every model to be nested under the new Community model.

## Community revamp

You can review the bulk of the community re-structure on these commits:

https://github.com/Terrastories/terrastories/pull/572/files/399ef976f523ac1f5520bc94b7ac33ef78a306d9

And the adjustments made for:

- [pundit authorization rescue for admin views](https://github.com/Terrastories/terrastories/pull/572/commits/f36272f61d8649696019f6aa1b79a92b6e19e1b6) (sans the navigation partial))
- [pundit create/update policy fixes](https://github.com/Terrastories/terrastories/pull/572/commits/3dc4cabee647b22083840249c3df91a04b75e20e)
- [pundit adjustment to ensure that index views are also properly authorized](https://github.com/Terrastories/terrastories/pull/572/commits/80e7e9940dea1a8edaf980578cc7ab359d363de6)
- [super admin introduction](https://github.com/Terrastories/terrastories/pull/572/files/12c29fe31d3dafdc7858578d29192b092cda00c1..5882e19ca5eed71f110553abd7537d2af8821946) which also includes view changes; but you'll get the gist.

This should help reviewing the monstrosity that is this PR given that we are working on a deadline for a demo.

## View / Administrate overhaul

The views are LARGELY the generated ones from Administrate, with some notable exceptions: 

- Brand new layout that overrides the default administrate sidebar / main area into a top navigation version instead.
- Reduced index views with an override to include place/speaker/user photos in the first column without a header, with the rest of the column attributes after. These currently still use the dashboards that Administrate offers; but could be updated to ignore them.
- Update show views to create "profile" like pages that highlight the important information; rather than the dl lists. While some areas still utilize this, it is, for the most part, gone now. I hope to remove the rest at some point, but would like to have a designer come through and reimagine what these look like better than I can do.


rails/Gemfile
```diff
@@ -77,6 +77,9 @@ group :development do
   # Add better error handling to make debugging simpler
   gem "better_errors"
   gem "binding_of_caller"
+
+  # annotate models
+  gem "annotate"
```
FeminismIsAwesome, line 82: lol maybe couldve done the annotate seperately but oh well

rails/app/assets/stylesheets/admin.scss
```diff
@@ -1 +1,4 @@
-@import 'admin/admin';
\ No newline at end of file
+@import 'core/media';
+@import 'core/colors';
+@import 'admin/admin';
+@import 'font-awesome';
```
FeminismIsAwesome, line 4: ditto re: styling changes

rails/app/controllers/admin/themes_controller.rb
```diff
@@ -1,36 +1,5 @@
...
-        deactivate_prev_theme
-      end
-      super
-    end
 
-    def update
-      if theme_params[:active] == "1"
-        deactivate_prev_theme
-      end
-      super
```
FeminismIsAwesome, line 14: context here? is there a way to preserve this work?

rails/app/views/admin/places/_collection.html.erb
```diff
@@ -0,0 +1,110 @@
...
+      <tr class="js-table-row"
+          <% if show_action? :show, resource %>
+            <%= %(tabindex=0 role=link data-url=#{polymorphic_path([namespace, resource])}) %>
+          <% end %>
+          >
+        <% collection_presenter.attributes_for(resource).each do |attribute| %>
+          <% if ["photo"].include?(attribute.name) %>
+            <td class="cell-data cell-data--<%= attribute.html_class %>">
+              <% if resource.photo.attached? %>
+                <%= image_tag url_for(resource.photo), height: "50", width: "50" %>
```
FeminismIsAwesome, line 69: what is different in this _collection.html.erb from the normal boilerplate?

rails/app/views/admin/places/show.html.erb
```diff
@@ -0,0 +1,91 @@
...
+    </dd>
+    <dt class="attribute-label">
+      <%= t("helpers.label.#{page.resource_name}.#{:description}") %>
+    </dt>
+    <dd class="attribute-data">
+      <%= page.resource.description %>
+    </dd>
+    <dt class="attribute-label">
+      <%= t("helpers.label.#{page.resource_name}.#{:region}") %>
+    </dt>
```
FeminismIsAwesome, line 57: haha this is why i hate adminstrate. carry on!


(like, the problems it solves, so much easier for barrier to entry with CRUD vanilla rails)

rails/app/controllers/home_controller.rb
```diff
@@ -3,6 +3,11 @@ class HomeController < ApplicationController
   def index
   end
 
+  def community_search_index
```
FeminismIsAwesome, line 6: gosh. so this just gives us the skeleton, and doesn't quite get to switching between or setting up home controller right for switching between communities?


Upon reading of the code (i read the rb and skipped the html, which appeared to be mostly styling changes)
This makes me concerned that if we had a deployment already out there, it would mess that up. Am i misreading that or ...? (because users who are not superusers would not have community ids associated with them)

rails/app/views/admin/places/_collection.html.erb
lauramosher: It hides a header item’s text a displays an image. Photo/media/etc

rails/app/controllers/admin/themes_controller.rb
lauramosher: A community only has one theme; most of this was so multiple themes could exist and get swapped which is no longer the case

rails/app/controllers/home_controller.rb
lauramosher: It does switch communities. A user can’t belong to more than one community so it doesn’t need to switch. Everything works as it did previously for a single user in the system... so long as that user belongs to a community. Which given how we set everything up each time, that should be the case.

As for the case where a user does not have a community, then it does get a little gnarly. We hope to have search setup in the future that allows users not associated with a community to search and see publicly available stories on a map.

All super users do is push you into a special admin dashboard for creating communities. They actually won’t have communities. Regular users will, which get directed as normal, and then users without a community get shoved into the unfinished search.

I highly recommend checking out the deployment to see it in action

rudokemper (approved):

PR #573, opened 2020-12-17
Attachable Theme background_img

lauramosher:
Instead of a db-level string for denoting a URL for where an image is, this allows an image to be uploaded and stored in active storage.

This also updates the seeds to create the Theme when the Community is created. An empty theme is created for a community, if one doesn't exist yet.


rudokemper (approved):

PR #574, opened 2020-12-17
Fix: New resource failed scopes

lauramosher:
This fixes two issues relating to creating stories/places/speakers in multi-instance:

1. Creating a new record wasn't properly relating to the logged in community. Import worked properly; but not through the regular CMS.
2. Speakers/Places/Stories associations were displaying *all* of the records, not just the ones scoped to the current community.

## Changes

* added adminstrate scoped has many and scoped belongs to gems
* overwrites new/create actions in speakers/places/stories controller to ensure community is properly attached (unnecesssary in update/edit since community should exist)

## Notes

This isn't ideal and I'm leaning toward trying to find a more flexible solution to support multi-instance. We're starting to hit the boundaries of what administrate is capable of doing with multi-instance and complex permissions without it getting more confusing.

For now, this will do so we can demo successfully and show that the multi-instance works as expected.


rudokemper (approved):

PR #576, opened 2020-12-17
Fix: permissions issue

lauramosher:
Editors/Admins could edit their own role; which caused them to lose the
privileges (or gain privileges) they weren't allowed to have.

Furthermore, super admin is for internal use only, and so that role
should be removed from the list of role options; this is done by adding
options :skip to skip keys in the enum


rudokemper (approved):

PR #577, opened 2020-12-18
Customize Mapbox Map Style and Access Token in Theme

lauramosher:
This adds the ability for admins to update the Theme to a custom mapbox map and access token, allowing a community to customize the style of their specific map view.

## Changes

* Adds db migration to add mapbox_style_url and mapbox_access_token to Theme table
* Updates theme administrate dashboard to allow editing for mapbox attributes ☝️ 
* Uses helper methods to switch between offline (local) mode, theme mapbox styles, and environmental variables fallback.

## Notes

# Fallback
image

# Theme set
image

### DIFF
--- a/rails/app/controllers/home_controller.rb
+++ b/rails/app/controllers/home_controller.rb
@@ -23,4 +23,20 @@ class HomeController < ApplicationController
   helper_method def stories
     policy_scope(Story)
   end
+
+  helper_method def mapbox_token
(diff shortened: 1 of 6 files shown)


rudokemper (approved):
QA tested and LGTM! We'll just have to file an issue to add some validation code later to ensure a super won't get away with only putting in a map and no token, and vice versa.

PR #590, opened 2020-12-24
Updating README and documentation to reflect new work on multi-instance and Themes

rudokemper:
This pull request updates the readme and documentation to reflect some of the latest work on multi-instance and Themes, in addition to some other minor updates like adding a note about Mapbox.com resource usage.


lauramosher (approved):
I think this is a great first brush at the new architecture.

PR #591, opened 2020-12-29
Ensure flash messages are shown on Devise sessions/sign in page

### LINKED ISSUE #584: [Rails] Our login screen doesn't have invalid credentials messaging
**Describe the bug**
When you try to log in to Terrastories with incorrect credentials, the page refreshes (as expected) but there is no invalid credentials messaging.

**To Reproduce**
1. Go to the Terrastories login screen.
2. Try to log in with some random credentials.
3. See the below screenshot (no messaging)

**Expected behavior**
Standard invalid credentials messaging from devise shows.

**Screenshots**
image

lauramosher:
Fixes #584

## Issue

The sessions new page utilizes flash messaging to display errors. Since the layout for devise does not include flash messaging, these messages were never shown to the user.

## Fix

At first, I added them to the layout; but noticed that the flash messages ended up being way far away from the focal point of the form and what went wrong. Instead, I added them directly to the sign in form area to keep the messaging near where their eyes already are.

image

### DIFF
--- a/rails/app/assets/stylesheets/components/devise.scss
+++ b/rails/app/assets/stylesheets/components/devise.scss
@@ -8,6 +8,18 @@
   align-items: center;
   text-align: center;
 
+  .flashes {
+    margin-bottom: 1rem;
+
+    .flash-alert {
+      color: $dark-red;
+    }
+
+    .flash-notice {
+      color: $dark-orange;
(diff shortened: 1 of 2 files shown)


rudokemper (approved):
LGTM

PR #595, opened 2021-01-02
Add Javascript testing for IntroPopup

jmk-198xtsuga:
Added a test for IntroPopup that uses Enzyme for element rendering.

It took a while to figure out all the React patterns and how to use Enzyme, but now I believe I have a thorough test for the IntroPopup component.

This sets up as a suite of Jest tests and then tests the following:

1. IntroPopup renders correctly as html
2. The IntroPopup text uses I18n translation
3. The IntroPopup initially opens popped up, and then closes when the close text is clicked
4. The IntroPopup defaults to popped up without any React.JS element props given
5. (optionally) The IntroPopup recognizes a "isPopped" prop for its initial behavior

One test checks that the shallow render matches a stored snapshot, which I think is the most efficient way to check its basic rendering, as this will ensure it has the right elements, structure, and content; and is easily reviewed and updated if there are future design changes.  Also, by using shallow it would not fail as the result of any inner components used (say if you added a video component at some point)

There are also checks that the component uses the supplied props as expected, which presently only includes the single isPopped prop.  Checks are done for the default behavior and both true and false isPopped states, though the code currently doesn't recognize the isPopped prop.  Review this, this could either be added as a feature or dropped from the test, I added this in as part of my exploration of how to test these elements.

Also included is a mock of I18n translation and a test that the rendering is done via I18n.  The mock includes a mock implementation to assist with shallow rendering inspection that identifies what content would be rendered for a certain translation.  This may make snapshots more intuitive for review into the future, and you are welcome to re-use.  I'm not sure the most appropriate place to put a test-only mocked implementation file, so I used a __mocks__ directory in the /rails/spec/javascript/components/ path.

If you're attentive to notice my commit history includes some extra commits for merges from different times I had my github and local repo catch up to your working repo so that I wouldn't eventually end up working way off into the wild.  Functionally there's only 1 new commit on mine (ad37eab) that included any productive work, so if you prefer commit history cleanliness you're not missing anything by squashing.

Also if you are attentive to the beforeAll / beforeEach designation, if I've read correctly beforeAll is run _once_ before any tests are run, whereas beforeEach is run multiple times to include once before each inner test (describe or it).  The tests at the main level of this file are designed to analyze one component ("component") without changing it, and the tests that cause changes each set up their own component individually, so I think beforeAll is a better pick here.

I am executing this test file by navigating to /rails/, running npm install (if needed) to ensure the javascript elements are in place, and then running:
`npm run-script test ./spec/javascript/components/IntroPopup.spec.js`

(or to test all javascript tests: `npm run-script test`)

This PR is supplied to help with issue #512 and may close the issue.

Let me know if you have any questions!


FeminismIsAwesome:
looks great! i appreciate the mocks for the locales, that makes sense to me.

FeminismIsAwesome (approved):

Issue #592, opened 2020-12-29
[Rails][Imports] Move import form from the bottom of the list views to a new view


lauramosher:
Thanks for adding an issue to Terrastories! Reach out to us in Slack or at terrastories-stewards@googlegroups.com if you have any questions about creating an issue.

## Context:

Importing Stories, Places, and Speakers is currently done by an inline form at the *bottom* of the respective index/list pages. This works well enough when there are no or very few items in the list, but can quickly get buried when there are many.

Let's make it so the import form is it's own view. This can be done by either making a new page view, or introducing a modal view of some kind.

## Acceptance criteria:

- [ ] Index/list pages should have a new button called "Import" that redirects to the new import form (modal or page).
- [ ] A new Import view that accepts the same file type and submits to the same route
- [ ] Once created, it redirects back to the index page.

## Assets

This could be worked in conjunction with #589.

The form at the bottom of this image is what should be moved into its own view.

![image](https://user-images.githubusercontent.com/31662219/103098149-b7457600-45d7-11eb-9b81-2e2e888d7553.png)

Aliciawyse:
👋🏾 hello there! I'd like to take this one :-)

Aliciawyse:
This new view should live here correct? https://github.com/Terrastories/terrastories/tree/master/rails/app/views/admin

PR #596, opened 2021-01-11
Create new view for importing

Aliciawyse:
### Issue
Fix for https://github.com/Terrastories/terrastories/issues/592

### Questions
- What should the headers be in the sample csv? Asking because when you view these pages in the app (`http://localhost:3000/admin/places?locale=en` and `http://localhost:3000/admin/speakers?locale=en`), you'll see a table with only a few aspects of places or speakers displayed.

- Is it a problem that I can successfully upload a csv for places with only name, description and region as the csv headers?

- This JS test keeps failing. I'm not sure how to confirm if this failure is caused by my changes. I think it's because I removed a `section` from the index/list admin page of speakers, places and stories
```
 FAIL  spec/javascript/components/Card.spec.js (11.449s)
  ● Card component › Given a user props › With an editor role › Renders the correct information

    expect(jest.fn()).toHaveBeenCalledTimes(2)

    Expected mock function to have been called two times, but it was called one time.

      28 | 
      29 |         expect(wrapper.find('.card--tasks').childAt(0)).toHaveLength(1);
    > 30 |         expect(global.I18n.currentLocale).toHaveBeenCalledTimes(2);
```

### DIFF
--- a/rails/app/controllers/admin/places_controller.rb
+++ b/rails/app/controllers/admin/places_controller.rb
@@ -52,9 +52,17 @@ module Admin
       else
         filepath = params[:file].read
         errors = Place.import_csv(filepath, current_community)
-        errors.empty? ? flash[:notice] = "Places were imported successfully!" : flash[:error] = errors
+        errors.empty? ? flash[:notice] = "Places were imported successfully! View them #{view_context.link_to 'here', admin_places_path}." : flash[:error] = errors
         redirect_back(fallback_location: root_path)
       end
+    end
+
+    def import_page
+      render "import_page"
+    end
+
+    def export_sample_csv
+      send_data Place.export_sample_csv, filename: "sample_places.csv"
     end
   end
 end
--- a/rails/app/controllers/admin/speakers_controller.rb
+++ b/rails/app/controllers/admin/speakers_controller.rb
@@ -69,7 +69,7 @@ module Admin
     end
 
     def export_sample_csv
-      send_data Speaker.export_sample_csv, filename: "sample_places.csv"
+      send_data Speaker.export_sample_csv, filename: "sample_speakers.csv"
     end
   end
 end
--- a/rails/app/controllers/admin/speakers_controller.rb
+++ b/rails/app/controllers/admin/speakers_controller.rb
@@ -51,7 +51,7 @@ module Admin
       else
         filepath = params[:file].read
         errors = Speaker.import_csv(filepath, current_community)
-        errors.empty? ? flash[:notice] = "Speakers were imported successfully!" : flash[:error] = errors
+        errors.empty? ? flash[:notice] = "Speakers were imported successfully! View them #{view_context.link_to 'here', admin_speakers_path}." : flash[:error] = errors
         redirect_back(fallback_location: root_path)
       end
     end
@@ -62,6 +62,14 @@ module Admin
 
     def default_sorting_direction
       :asc
+    end
+
+    def import_page
+      render "import_page"
+    end
+
+    def export_sample_csv
+      send_data Speaker.export_sample_csv, filename: "sample_places.csv"
     end
   end
 end
--- a/rails/app/views/admin/stories/import_page.html.erb
+++ b/rails/app/views/admin/stories/import_page.html.erb
@@ -1,0 +1,23 @@
+<section class="main-content__body padding-fix color-fix">
+  <p>
+    IMPORTANT: Need to import stories? The order of the values in your csv must be in a particular order. Feel free to use this <%= link_to "SAMPLE CSV", export_sample_csv_admin_stories_path %> to get started with successfully uploading stories.
+  </p>
+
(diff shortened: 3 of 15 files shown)


rails/app/views/admin/stories/import_page.html.erb
```diff
@@ -0,0 +1,23 @@
+<section class="main-content__body padding-fix color-fix">
+  <p>
+    IMPORTANT: Need to import stories? The order of the values in your csv must be in a particular order. Feel free to use this <%= link_to "SAMPLE CSV", export_sample_csv_admin_stories_path %> to get started with successfully uploading stories.
+  </p>
+
+  <%= form_tag send("import_csv_admin_stories_path"), multipart: true do %>
```
Aliciawyse, line 6: Something feels weird about this change from `import_csv_admin_#{page.resource_name.pluralize}_path` to `import_csv_admin_stories_path`

FeminismIsAwesome: @Aliciawyse i think thats totally fine! We're moving from a generic form to forms that are more specific to the resource's need. I've found the administrate abstraction to usually get in the way anyway; anything thats more specific and rails-y feels a little easier to read too. 

If we do end up wanting to generecize it again, we could always find the common pieces between the forms and make a partial that goes back to the `import_csv_admin_#{page.resource_name.pluralize}_path` style.

lauramosher: @Aliciawyse I agree with @FeminismIsAwesome in this; the form tag itself is going to be "duplicated" across each specialized view. The Administrate "way" forced us to do weird things like interpolating the resource into the path and I for one am glad to see that go.

IF we wanted to continue to share some of the logic, we could create a partial view for the form; but I'm okay with waiting on this until we see the repetition.

rails/app/controllers/admin/speakers_controller.rb
```diff
@@ -63,5 +63,13 @@ def default_sorting_attribute
     def default_sorting_direction
       :asc
     end
+
+    def import_page
+      render "import_page"
+    end
+
+    def export_sample_csv
+      send_data Speaker.export_sample_csv, filename: "sample_places.csv"
```
FeminismIsAwesome, line 72: @Aliciawyse i think you meant sample_speakers.csv! :D

rails/spec/models/speaker_spec.rb
```diff
@@ -136,4 +136,10 @@
       it { is_expected.to be_nil }
     end
   end
+
+  describe 'export_sample_csv' do
+    it 'downloads a csv' do
+      expect(described_class.export_sample_csv).to eq("name,birthdate,community\n")
```
FeminismIsAwesome, line 142: speakers are missing the media column

rails/spec/models/place_spec.rb
```diff
@@ -133,4 +133,10 @@
       expect(place.point_geojson.keys).to include('properties')
     end
   end
+
+  describe 'export_sample_csv' do
+    it 'downloads a csv' do
+      expect(described_class.export_sample_csv).to eq("name,description,region\n")
```
FeminismIsAwesome, line 139: places also have a lat, long, and media.

FeminismIsAwesome:
@Aliciawyse some replies to your questions):

1) try adding the columns i suggested to the default csvs (it was missing a few, which is fair it was undocumented so hard to spot, i just looked at FileImport::PlaceRowDecorator and the Speaker code to see what was up when it failed to import my speakers). That hopefully should convert the nil issues to blanks, but even if they persist, I'd accept this code and then make a follow up feature to not bug out if media files are missing on speaker/story. You can see the bug on line 23 of the CsvHelper class.

2) that's a great question re: the js tests. I thought it was weird, so I checked myself by moving from your branch to master, which should have none of your changes, and seeing the same test fail. So I think we just accidentally merged in a failing test, and I'll add a ticket to fix that.

Let me know if you have further questions!

rails/app/controllers/admin/places_controller.rb
```diff
@@ -52,9 +52,17 @@ def import_csv
...
         filepath = params[:file].read
         errors = Place.import_csv(filepath, current_community)
-        errors.empty? ? flash[:notice] = "Places were imported successfully!" : flash[:error] = errors
+        errors.empty? ? flash[:notice] = "Places were imported successfully! View them #{view_context.link_to 'here', admin_places_path}." : flash[:error] = errors
         redirect_back(fallback_location: root_path)
       end
     end
+
+    def import_page
+      render "import_page"
```
FeminismIsAwesome, line 61: @Aliciawyse are those renders necessary? doesn't rails try to render a page with matching name by default?

FeminismIsAwesome (approved):
had one more question re: a render call, but honestly lgtm

Aliciawyse:
oh good question. I'll double check that and follow up

PR #602, opened 2021-02-02
Destroy associations on deletion.

lauramosher:
When a community is destroyed, destroy speakers, places, and stories
associated with the Community.

Nullify users relationship with the community.


lauramosher:
@rudokemper This will make sure when you delete communities, their associated records are also deleted. I guess we overlooked this before!

FeminismIsAwesome (approved):

Issue #600, opened 2021-02-01
[Bug][React] When a map Marker popup expands in size to accommodate an image, the X to close the popup stays where it was


rudokemper:
**Describe the bug**
In Terrastories, it is possible to add an image for a Place marker. This gets visualized in the marker popup, when a user clicks on a Place marker point on the map.

If there is an image for a place, the popup expands in size to accommodate the image. But the X to close the popup stays where it was, in the original popup dimensions.

**To Reproduce**
1. Add an image for a place.
2. Go to the map and click on the place to open the popup.
3.  See how the popup first loads as though there was no image, it then expands to include the image, but the X will stay where it was before the expansion.

**Expected behavior**
When the popup expands, the X to close the popup also adjusts to be in the top right of the popup.

**Screenshots**
![image](https://user-images.githubusercontent.com/31662219/106526868-594c5e00-64b4-11eb-8799-cf2e3bd0007f.png)

astojanowski:
Hi! I have fixed this bug.

rudokemper:
Thanks so much @astojanowski!

astojanowski:
I'm glad to help :)

PR #607, opened 2021-02-15
Fix X button in map Marker containing an image (#600)

astojanowski:
Hi! This is a fix for #600. I tested it for markers with and without images, both long and short names (issue exists for short names only).


FeminismIsAwesome (approved):
lgtm!

rudokemper:
same here. Thanks!

Issue #597, opened 2021-01-31
[Bug][Map] Sidebar filters dropdown are not filtering Places on the map.


rudokemper:
**Describe the bug**
The filter dropdowns in the Terrastories sidebar are supposed to do a few things: 

Filter the stories in the sidebar,
Zoom to the maximum extent of the Places with filtered stories,
Hide Place markers on the map that don't have a story in the active filter.

Currently, this last step is not working. 

**To Reproduce**
Steps to reproduce the behavior (this is one example):
1. Use one of the dropdown filters.
2. It will zoom to the maximum extent of the Place markers.
3. Navigate the map to where there were other Place markers that should have been filtered out. You will still see them show up.

**Expected behavior**
The only Place markers on the map that should show up are Places that have a Story in the sidebar. (I.e. if there are 100 places in Terrastories, and an active filter is only showing stories for 25 places, only those 25 Place markers are shown).

**Additional context**
This demo, running a November 2020 version of the Terrastories codebase, does have the behavior as desired working. http://terrastories-2020imw.herokuapp.com/

rudokemper:
@lauramosher I think this may have come up during the December Tech Matters work. Because the filters were operating just fine in November.

lauramosher:
@rudokemper I looked into this — the filter logic is all encompassed in the React code; not the rails code. This was likely accidentally introduced when we merged the new Map + Marker components since it changed how we update the markers on filtering.

See: https://github.com/Terrastories/terrastories/pull/563

Updating the labels. @mirandawang Do you have time to look into this?

rudokemper:
Yeah, my mistake on this -- I thought that the November 2020 Heroku demo above was using a version of the codebase with the map marker refactoring work already done, but that is not the case. That took place shortly after IMW and before the TM work. Was getting my timelines confused :)

rudokemper:
That's great! Thanks @astojanowski I just assigned it to you.

astojanowski:
Fixed by updating map markers source every time the list of places has changed. Additionally I removed duplicates from the list of points, couldn't find any use of these duplicates...

PR #611, opened 2021-03-03
Close marker popup when marker is filtered out

### LINKED ISSUE #474: [Map] Marker popup doesn't close when the marker disappears from the map
When you click on a marker, a popup opens. But when you make that marker disappears i.e. by adjusting the filters, the popup stays open.

image

Let's make it so that the popup closes automatically whenever the marker disappears from the map.

This issue can be observed on the Heroku demo.

astojanowski:
Fixes #474

There's a similiar issue #597 for markers (not popups) not disappearing after filtering - I'm going to work on this too.

### DIFF
--- a/rails/app/javascript/components/App.jsx
+++ b/rails/app/javascript/components/App.jsx
@@ -156,10 +156,17 @@ class App extends Component {
         bounds,
         maxZoom: 12
       };
+
+      var activePoint = this.state.activePoint;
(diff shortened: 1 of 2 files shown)


lauramosher:
Thank you so much for your contribution! We will review and test this out soon.

rudokemper (approved):
Looks great! Approved 👍

Issue #601, opened 2021-02-02
[Bug][Rails] CSV Importer should update existing entries instead of creating duplicate ones


rudokemper:
**Describe the bug**

**To Reproduce**
[CSV.zip](https://github.com/Terrastories/terrastories/files/5907717/CSV.zip)

1. Import the Speaker CSV in this zip that has some places in it for `Birthplace`. 

2. Import the Places CSV in this zip, which has the same places (such as Bosulanti, Bilawata in the example below).

3. You will see that instead of updating Bosulanti and Bilawata, it just creates 

**Expected behavior**
The CSV importer should check to see if there is already an existing Place with the same name. If so, it can update the content of that Place, and not create a duplicate Place.

**Screenshots**
![image](https://user-images.githubusercontent.com/31662219/106534213-92d79600-64c1-11eb-8fb8-670f99bf5a29.png)

PR #614, opened 2021-03-23
[Fix] CSV file import

GQuirino:
Issue #601 .
This PR fixes a bug which was duplicating places instead of just updating its info when place exists.


lauramosher (approved):
Thank you!

Nice test, too.

Issue #587, opened 2020-12-24
[Custom themes] Let's make it possible to set the map center, extent, zoom, pitch, and bearing for a community in Theme


rudokemper:
Now that we have a Theme model in Rails, let's make it possible to set the default map center, boundaries, zoom, pitch, and bearing for a community. Currently, these are hard coded in `Map.jsx`.

## Acceptance criteria:
- [ ] Admin user logs into Terrastories and goes to the Theme menu.
- [ ] There are fields for 
    * center latitude and longitude
    * southwest boundary latitude and longitude
    * northeast boundary latitude and longitude
    * zoom level
    * pitch
    * bearing
- [ ] Entering values for these overrides whatever is set in `Map.jsx`.
- [ ] Add validation tests for each field:
    * Center `latitude` is between -90 and 90, `longitude` is between -180 and 180.
    * Southwest boundary `latitude` is between -90 and 90, `longitude` is between -180 and 180.
    * Northeast boundary `latitude` is between -90 and 90, `longitude` is between -180 and 180.
    * Zoom is between 0 and 22.
    * Pitch is between 0 and 85.
    * Bearing is between -180 and 180.

maxkadel:
Interested in working on this one! 

(also, if you put spaces between the []s you can use the checkmarks)

maxkadel:
If these values aren't set in the theme, what should the defaults be? The current defaults work well with the seed set, which is set to the 2 US coasts. 

![image](https://user-images.githubusercontent.com/45948126/112316292-59c9cf80-8c81-11eb-9215-68f00ea313ad.png)

lauramosher:
@rudokemper Do you have thoughts on this?

rudokemper:
Yep, because of the seed set I think the current defaults are fine. Thanks @maxkadel.

PR #615, opened 2021-03-24
587 map config theme

maxkadel:
- Make theme available from Home controller, making it so we don't have to make multiple database calls for one object
- Add map defaults to theme & theme administration dashboard
  - Broke out the points into separate latitude and longitude to match the current "places" modeling
  - Pulling the json into the javascript, used the string type for the points, because it allowed for keeping the level of precision I think we want
  - I wanted to just replace the defaults in `rails/app/javascript/components/Map.jsx`, but I wasn't able to figure out how to do so (I don't have much experience with js), and I thought working was a good place to start. Commented the constants out for now in case someone has a clear idea of how to replace them from the database values
- Separated the map configuration into its own concern (MapConfigurable) which is then included in the theme, to make it clear it all goes together, and in case it ends up in a separate class eventually
  - This concern sets the defaults, if they aren't provided on theme creation
  - It also includes methods to return the constants as arrays that parallel the original constants in the javascript. If we don't end up using them we might want to remove these methods
- The model includes the requested validations
- Tests
  - Change the controller test to a request test - the controller test was not running originally, and had more luck with inspectability in the request spec
  - Add Warden test helpers to test login functions of the request spec
  - Added tests for the theme defaults and validations

Connected to https://github.com/Terrastories/terrastories/issues/587 

image
image


rails/app/javascript/components/Map.jsx
```diff
@@ -5,14 +5,14 @@ import MiniMap from "../vendor/mapboxgl-control-minimap";
...
-  [-180, -85], //southwest
-  [180, 85] //northeast
-];
-const defaultZoom = 3.5;
-const defaultPitch = 0;
-const defaultBearing = 0;
+// const defaultCenter = [-108, 38.5];
+// const defaultBounds = [
+//   [-180, -85], //southwest
+//   [180, 85] //northeast
```
FeminismIsAwesome, line 11: if w're not using this anymore, do we want to consider just pulling it out entirely?

rails/app/models/theme.rb
```diff
@@ -1,9 +1,19 @@
 class Theme < ApplicationRecord
+  include MapConfigurable
   has_one_attached :background_img
   has_many_attached :sponsor_logos
+  after_initialize :set_map_defaults
```
FeminismIsAwesome, line 5: i think this is an interesting way to avoid worrying about pre-existing values for now, and think it works so good for now.

It feels a little funny long term, but i might just be more attached to db defaults.

FeminismIsAwesome (approved):

PR #624, opened 2021-04-26
606 replace email with username

ChrisHall1:
Added username field to users table. Users log in with a username instead of email. When creating an account there is also a username field. Previous references to email in user dashboards replaced with username. Work still needs to be done to make email not required when creating an account, but that will take some consideration into how to manage an account without an email associated with it.

### DIFF
--- a/rails/app/controllers/application_controller.rb
+++ b/rails/app/controllers/application_controller.rb
@@ -13,6 +13,10 @@ class ApplicationController < ActionController::Base
   end
 
   private
+  
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:username)
+  end
 
   def set_locale
     I18n.locale = params[:locale] || I18n.default_locale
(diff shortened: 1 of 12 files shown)


rudokemper:
Hi @ChrisHall1, sorry that it's taken us a while to get back to you. Thanks for this PR; we did discuss this a few times, but didn't get the opportunity to write back just yet.

Also, we're thinking we didn't scope this out properly enough. Right now, there are production servers that are using email address to log in, and this code isn't providing any backwards compatibility towards that. Moreover, we discussed this and believe that we should allow for users to log in with _either_ email address _or_ username. That way, online users can still use email (and receive password reset instructions in the future if they are locked out), and offline users can access with username. 

We probably need to do a bit more scoping, but we can use your PR to progress towards this, although we can't ship it in its current state. If you're interested in tackling this more robust feature (of permitting logins with either email or username), that would definitely be welcome too!

Again, our bad on dropping the ball on responding to this PR in a timely fashion, but your work is very much appreciated.

rudokemper:
Archiving this work, but closing the PR for now.

Issue #588, opened 2020-12-24
[Rails] Create a "topic" field for story and add it to Administrate views and sidebar filter


rudokemper:
## Context:
Some communities have told us that they would like to see a "Topic" field to be able to filter through stories. Let's 

1) do a migration to add this field to the `Story` model
2) add it to the Administrate views for viewing and editing stories (as a typeahead field).
3) add it as a filter option in the first dropdown on the map view.

## Acceptance criteria:
[] When an admin user logs in to the Administrate back end, and opens a story, `Topics` is shown as a block on the sidebar (similar to `Description`).
[] When the admin user edits a story, `Topics` shows up as a typeahead field under `Language`. 
[] When any user visits the map, `Topics` is one of the dropdown options in the first dropdown of `Filter stories` if there are stories with a non-empty topic field. 
[] The second dropdown will be populated with the values provided for `Topics` in the stories. When selecting a topic, the sidebar and map filters the stories & associated markers, as per the standard filtering behavior.
[] Existing tests do not break

julinvictus:
I can give it a try to this one

julinvictus:
Done✅

PR #625, opened 2021-04-30
Add topic to story

julinvictus:
Issue #588 

This PR includes:
- Migration to add `topic` field for `stories` table
- Add `topic` to Administrate views
- Add `topic` to localization files
- Add `topic` to sidebar filter in UI (WIP)

Screen Shot 2021-04-29 at 8 39 14 PM

Screen Shot 2021-05-08 at 4 16 57 PM

### DIFF
--- a/rails/app/javascript/components/App.jsx
+++ b/rails/app/javascript/components/App.jsx
@@ -100,6 +100,20 @@ class App extends Component {
           filterMap[category] = Array.from(speakerSet).sort();
           break;
         }
+        case FILTER_CATEGORIES[3]: {
+          // second category: Topic
+          const topicSet = new Set(
+            this.props.stories
+              .map(story => {
+                return story.points.map(
+                  point => point.properties.topic
+                );
+              })
+              .flat()
+          );
+          filterMap[category] = Array.from(topicSet).sort();
+          break;
+        }
       }
     });
     return filterMap;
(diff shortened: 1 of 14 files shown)


rails/app/javascript/components/App.jsx
```diff
@@ -100,6 +100,20 @@ class App extends Component {
           filterMap[category] = Array.from(speakerSet).sort();
           break;
         }
+        case FILTER_CATEGORIES[3]: {
+          // second category: Topic
+          const topicSet = new Set(
+            this.props.stories
+              .map(story => {
+                return story.points.map(
```
FeminismIsAwesome, line 108: @julinvictus 
2 things
1) you need to add the JSON field of topic to 
terrastories/rails/app/views/home/_home.json.jbuilder , probably in a similar way to the json.extract! story on line 2.
2) when iterating through the stories, i think story.points would be the places on the map, i think for a story field you coould just directly grab that from story.topic, once you do add it to the home json.

FeminismIsAwesome: also, not necessary but a jest test could help drive some of that relationship

julinvictus:
Ok now I can see topic options in the dropdown menu but when I click on it, it breaks the app. I tweaked some of that _home.json.jbuilder and react files but couldn't figure out.

rails/app/javascript/components/App.jsx
```diff
@@ -157,6 +171,23 @@ class App extends Component {
         });
         break;
       }
+      case FILTER_CATEGORIES[3]: {
+        // second category: topic
+        filteredStories = this.props.stories.filter(story => {
+          if (
+            story.points.some(point => {
+              return (
```
FeminismIsAwesome, line 175: @julinvictus you fixed the category dropdown, but the filter now has the same problem:

this filter just wants to be "if i am filtering by topic, return me all stories that meet this filter

and so theres no need to iterate through points, since we're just looking at the top level story'

something akin to
  filteredStories = this.props.stories.filter(story => {
          if(story.topic) {
            return story.topic.toLowerCase() == item.toLowerCase();
          }
        });

FeminismIsAwesome (approved):
lgtm

PR #633, opened 2021-06-26
564 bugfix for browser crash when clearing filter

Julien-Bidar:
I tried a simple approach which is to set back the itempsOptions to an empty array when the filters fields are cleared.


rudokemper (approved):
Tried it out and this successfully solves the problem! Thanks so much!

Issue #646, opened 2021-09-10
[Rails] Update dashboard columns for Stories and Places


rudokemper:
## Context:
In the Terrastories administrative menu, we have dashboards for Speakers, Stories, and Places from which you can view, edit, import, add, and delete data for each.

Currently, there are some fields that don't need to be shown for the Places and Stories dashboard views, and some that are missing. Let's have the fields be the ones in the acceptance criteria below (and in the order listed).

## Acceptance criteria:

STORIES

- [ ] `Title` - this one is good as is.
- [ ] `Description ` - let's add this here.
- [ ] `Speakers` - let's add this here.
- [ ] `Places` - let's add this here.
- [ ] `Permission level` - let's add this here.

`Language`, `Interview location` and `Date interviewed` should be removed.

PLACES
- [ ] `Photo` - this one is good as is
- [ ] `Name` - this one is good as is
- [ ] `Description` - this one is good as is
- [ ] `Type of Place` - let's add this here
- [ ] `Region` - this one is good as is

julinvictus:
Hi I can work on this one

PR #662, opened 2021-10-05
Update dashboard columns

julinvictus:
Issue #646 

This PR includes:
Update in dashboard columns for Stories and Places

Screen Shot 2021-10-04 at 10 25 17 PM
Screen Shot 2021-10-04 at 10 24 43 PM


lauramosher:
@rudokemper This is fine code-wise with what Administrate provides by default with this configuration, so I've marked it as hacktoberfest-accepted. I'm going to leave final approval up to you in case you had something specific in mind for how the information is displayed in each row.

rudokemper:
Hi @julinvictus! Thanks for this one. Is there any way that we can show the names of the speakers and places, instead of indicating a number `x speakers/places`?

So for "Terrastories Team testimonial", under "Speakers", it would show the four speakers, separated by a comma. 

We may consequently want to adapt the width of each of the columns, but that can be a separate ticket.

rudokemper:
Looking really good, @julinvictus !

One minor thing; on my end, it looks like the Places field titles are showing the system field names rather than the desired output. Do you see that as well?

image

Issue #651, opened 2021-09-10
[Rails][Administrate] Update Theme form labels to be more user-friendly 


rudokemper:
## Context:
Currently, the Theme field labels are not the most intuitive. Let's change these to the following

## Acceptance criteria:

- [ ] `Background img` - Background image (for welcome screen)
- [ ] `Sponsor logos` - Sponsor logos (for welcome screen)
- [ ] `Mapbox style url` - Mapbox style URL (for online maps)
- [ ] `Mapbox access token` - Mapbox access token associated with the style
- [ ] `Center lat` - Map center, latitude
- [ ] `Center long`- Map center, longitude
- [ ] `Sw boundary lat` - SW bounding box, latitude
- [ ] `Sw boundary long` - SW bounding box, longitude
- [ ] `Ne boundary lat` - NE bounding box, latitude
- [ ] `Ne boundary long` - NE bounding box, longitude
- [ ] `Zoom` - Map zoom level (1 to 22)
- [ ] `Pitch` - Map pitch degree (0 to 90)
- [ ] `Bearing` - Map bearing degree (-180 to 180)

## Assets
![image](https://user-images.githubusercontent.com/31662219/132916202-968f2e76-9f1e-4826-b778-7edb7e44afc5.png)

julinvictus:
Hi I can work on this one

PR #670, opened 2021-10-06
Update Theme form labels

julinvictus:
Issue #651 

This PR includes:
Update to Theme form labels
Screen Shot 2021-10-05 at 8 26 31 PM (2)


rudokemper (approved):
Looks good! We'll want to do some i18n when we're done with all these text changes, but that can be a separate ticket.

Issue #663, opened 2021-10-06
[Bug][Rails] Region typeahead field should only pull suggestions from active community


rudokemper:
**Describe the bug**
The `Region` field in the Places model is configured as a typeahead field, helpfully providing you with existing values when you add or edit this field. However, currently it is populating the list of values from _all_ of the communities on one Terrastories server, instead of just the one you are logged in to.

**To Reproduce**
1. Create or log in to a Terrastories community
2. Create some place data with the Region field filled in
3. Log out of this community, and create or log into a different Terrastories community
4. Create a place, and see the values from the first community in the suggestions for Region

**Expected behavior**
You should only see values for `Region` from the community you are logged in to.

**Screenshots**
![image](https://user-images.githubusercontent.com/31662219/136123891-20ea2496-3736-4154-bf83-d3ae1660ea25.png)
(These are all values from different communities residing on the same server).

ConnieJChi:
👋 Hello! Do you mind if I try working on this issue?

rudokemper:
Hello @ConnieJChi ! Welcome and thanks :)

PR #681, opened 2021-10-10
633: Adding proper filters for dropdown lists

ConnieJChi:
Referring to issue [633](https://github.com/Terrastories/terrastories/issues/663) where the suggested region and type of place for a new places entry and the suggested topic for a new story entry was not properly filtered by the user's community.


rudokemper:
I tested this PR out and can confirm it's accomplishing the desired behavior! 

But wanted to see if @lauramosher could take a quick look at the code, since she's worked on the multi-instance architecture extensively.

Issue #672, opened 2021-10-06
[Rails] Change the order of options in the primary filter to alphabetical


rudokemper:
Let's change the order of the options in the primary filter to be alphabetical. 

So with the current selections, it would be `Region`, `Topic`, `Type of Place`, `Speaker`. If there were to be one added in the future, it would automatically be placed alphabetically as well.

(Note that the `filterMap` and `handleFilter` arrays in `App.jsx` have to be adjusted as well).

julinvictus:
Hi I can work on this. Also I think you mean `Region, Speaker, Topic, Type of Place`

rudokemper:
Ah you are right :) thanks @julinvictus !

PR #682, opened 2021-10-10
Change primary filter to alphabetical order

julinvictus:
Issue #672 

This PR includes:
- Sorted array in `FilterConstants`
- Modified access to above array by adding proper element to switch case, instead of using array index. By doing this, order of cases in the switch won't matter.
Screen Shot 2021-10-10 at 10 30 49 AM


rudokemper:
Looks great @julinvictus ! Please feel free to rebase, then squash and merge 🙂

PR #692, opened 2021-10-13
Add introductory panel and remove pop up card

griseduardo:
**Motivation**:
Add introductory panel and remove right pop up card

**Solution**:

- Create Introductory Panel component
- Remove Pop Up card
- Create reusable Button component, to add 'EXPLORE' button
- Create logic to change Introductory Panel to show stories

**Issue**:
https://github.com/Terrastories/terrastories/issues/675

Opening page:
before-click-explore

After button click:
after-click-explore

PS: I updated all i18n languages using English

### DIFF
--- a/rails/app/javascript/components/StoryList.jsx
+++ b/rails/app/javascript/components/StoryList.jsx
@@ -94,7 +94,7 @@ class StoryList extends Component {
             itemOptions={this.props.itemOptions}
           />
         </div>
-          {!this.state.showStories
+          {!this.state.showStories && window.innerWidth > 600
             ?
           <div className="card--nav terrastory-info">
             <IntroductoryPanel 


rudokemper:
Hi @griseduardo, thanks for taking this one on. We haven't done a code review yet, but I just tested this PR out on the front and it looks really good!

I think there is just one additional change that we should make based on front-end testing: once a user selects an option in the second drop down, the intro panel should automatically disappear and the list of filtered cards revealed.

Otherwise, great work on this!

griseduardo:
> Hi @griseduardo, thanks for taking this one on. We haven't done a code review yet, but I just tested this PR out on the front and it looks really good!
> 
> I think there is just one additional change that we should make based on front-end testing: once a user selects an option in the second drop down, the intro panel should automatically disappear and the list of filtered cards revealed.
> 
> Otherwise, great work on this!

Hello @rudokemper , thank you.
I updated with the change you presented.

rudokemper:
Awesome! Looks great. One final thing I just noticed is that we may want to adapt to the mobile view. Would it be easy and possible to hide this panel for mobile, for the time being? (Mobile UI is due for a redesign, where we'll consider how to incorporate this panel differently.)

rudokemper:
Also, sorry I didn't catch your question about i18n on the other thread. I think what you did was fine; we have to do some i18n updates to a number of the other languages anyway.

griseduardo:
> Awesome! Looks great. One final thing I just noticed is that we may want to adapt to the mobile view. Would it be easy and possible to hide this panel for mobile, for the time being? (Mobile UI is due for a redesign, where we'll consider how to incorporate this panel differently.)

One question, do you want to hide the introductory panel and return the stories when in mobile? (like it was before my changes)

PR #693, opened 2021-10-15
Added partials for Theme's edit and show views and redefined display_…

EitanFuturo:
…resource method for ThemeDashboard. Also updated locals for translations needed.


EitanFuturo:
hi @lauramosher, the branch should be up to date now. Please tell me if it is not so.

PR #694, opened 2021-10-17
Adapt height of the welcome screen card to adjust to overflow content

### LINKED ISSUE #684: [Bug][Rails] Adapt height of the welcome screen card to adjust to overflow content
**Describe the bug**
Currently, if any of the content on the welcome screen card (such as the `h1` welcome message, `.welcome-card-user`, or `.language-picker` content) is forced to go to a second line, the overall `.welcome-card` height does not adjust. Let's make it dynamically adjust so that all the content fits.

**To Reproduce**
1. Go to the Terrastories welcome screen
2. Use DevTools to make the content of any of the above adapt to two lines.

**Expected behavior**
The height of the `.welcome-card` should increase to accommodate the content needing to increase in size.

**Screenshots**
image
In this screenshot, the `.welcome-card-user` content forces the `.language-picker` content down, and while it's still visible there should be a 2em padding at the bottom as on the other sides..

image
In this screenshot, both the `h1` and `.welcome-card-user` content are taking up two lines, and the `language-picker` block isn't even visible.

everly-gif:
Fixes #684 

### What I did: 

1. The height is no longer fixed , it dynamic changes as the screen size decreases, I have set a min-height property which also solves the overflowing content issue (no longer overflows anymore)
2. All the content of the card appears in a single line upto 355px. Added some media queries with dynamic font-sizing.

### Here's a screenshot for 437px (no overflow of content , no forced two lines anymore) : 
ss

### DIFF
--- a/rails/app/assets/stylesheets/components/welcome.scss
+++ b/rails/app/assets/stylesheets/components/welcome.scss
@@ -158,8 +158,4 @@
     font-size: 3.1vw;
   }
 }
-// @media screen and (max-width:450px){
-//   .welcome-card{
-//     min-height: 15rem;
-//   }
-// }
+


rudokemper:
@everly-gif thanks. I can confirm with your branch that the height is adapting to overflow content, both on desktop and mobile view. 

But there seems to be an issue with how the logo interacts with the new width properties. This is what I'm seeing on Firefox & Chrome, on a Windows machine.

image

Issue #689, opened 2021-10-11
[Rails] Add language field to Story cards in sidebar


rudokemper:
## Context:
We have a `language` field for the Story model, but we're not using it yet on the front end. Let's add it to the story cards on the Terrastories sidebar, per the acceptance criteria below.

## Acceptance criteria:

- [ ] If `language` is set for a story, add a `<p>` tag underneath the video with `**Language:** [value]` in the story card
- [ ] If `language` is not set for a story, do not add a `<p>` tag.
- [ ] Use `language` translation from the i18n locales in `config/locales`.

## Assets
![image](https://user-images.githubusercontent.com/31662219/136837914-cd193074-def0-4f59-b4c1-cd84e36f7b23.png)

PR #700, opened 2021-10-21
Issue #689: Add story field to Story cards in sidebar

noelledusahel:
- If Story is created and language specified, the user should see the language field in the story sidebar.
- If no language is specified, language field will not be displayed
- "Language" should translate according to I18n constants

Screen Shot 2021-10-21 at 11 18 45 AM


rudokemper:
Thanks so much @noelledusahel! This is great :)

Issue #688, opened 2021-10-11
[Rails] Let's add Community and Language fields as filter options; hide them if there are no values


rudokemper:
## Context:
There are two fields in Terrastories that we are currently not using on the front end: `community` (in the Speaker model) and `language` (in the Story model). These could be set as filters for the stories. 

Let's add these to the filters menu, but, make it so that they don't show up if there are no values set within the community. 

(That is, if there are no stories that have any content in `language`, Language does not show up as one of the filter menu options.)

## Acceptance criteria:

- [ ] User is able to filter by language.
  - [ ] Only stories marked as being in that language show in the sidebar.
  - [ ] Only places related to those stories in that language show on the map.
- [ ] User is able to filter by community.
  - [ ] Only stories contributed by a Speaker from that community show in the sidebar.
  - [ ] Only places related to those stories show on the map.
- [ ] If there are no stories available that have content in the Language field, then the Language option does not show up in the first filter menu
- [ ] If there are no speakers that have content in the Community field, then the Community option does not show up in the first filter menu
- [ ] BONUS: Adapt the same "hide filter menu option if no values are set" for `topic`

rudokemper:
Hi @noelledusahel ! Did you still want to work on this one? Just checking in :)

noelledusahel:
> Hi @noelledusahel ! Did you still want to work on this one? Just checking in :)

I put this PR up last October... I dont think it got a review since  https://github.com/Terrastories/terrastories/pull/702

rudokemper:
Oh! My bad @noelledusahel I didn't realize this PR also addressed this issue. We'll get it reviewed as soon as possible; our stewards team is a little small at the moment (and still mostly volunteer). Sorry about that.

PR #702, opened 2021-10-29
#688-Add Community and Language fields as filter options

noelledusahel:
## Context:

Issue https://github.com/Terrastories/terrastories/issues/688
There are two fields in Terrastories that we are currently not using on the front end: community (in the Speaker model) and language (in the Story model). These could be set as filters for the stories.

Let's add these to the filters menu, but, make it so that they don't show up if there are no values set within the community.

(That is, if there are no stories that have any content in language, Language does not show up as one of the filter menu options.)

## Acceptance criteria:
- [x] User is able to filter by language.
   - [x] Only stories marked as being in that language show in the sidebar.
   - [x] Only places related to those stories in that language show on the map.
- [x] User is able to filter by community.
  - [x] Only stories contributed by a Speaker from that community show in the sidebar.
  - [x] Only places related to those stories show on the map.
- [x] If there are no stories available that have content in the Language field, then the Language option does not show up in the first filter menu
- [x] If there are no speakers that have content in the Community field, then the Community option does not show up in the first filter menu
  - [x] BONUS: Adapt the same "hide filter menu option if no values are set" for topic
Screen Shot 2021-10-29 at 11 29 45 AM
Screen Shot 2021-10-29 at 11 29 56 AM


lauramosher:
Thanks so much for your contribution! We will review this shortly!

rudokemper (approved):
LGTM! Sorry for the delay.

PR #707, opened 2021-12-25
#676 Add Audio Field to Place

EitanFuturo:
## Changes
- Added name_audio relation, validation, and translation for Place model. 
- Added attribute to PlaceDashboard and element to view component.
- Added route, action, dashboard configuration for deleting audio attachments for places.

## Caveats
- Didn't add the “MP3 or WAV files only” message because it implies redefining administrate partials for this only purpose and didn't find the place that would only apply for this case.
- Didn't add the support for importing a Place with an audio file, which is also needed for updating the model tests. The reason is that it wasn't required and implies modifying different parts of the application including creating a new PlaceAudioDecorator. I'll submit a feature request issue with this.
- Found an error, which I didn't correct but will do after creating a new issue.
  - Currently, the remove link for photos added to a Place is throwing an error, and I would think it's the same for other models. For correcting this is necessary to add a route, action, and correct path to the destroy_path passed in options in the PlaceDashboard ATTRIBUTE_TYPES, as I did in this PR with name_audio attribute.


rudokemper:
This is great, @EitanFuturo! We'll take a close code review soon, but in advance wanted to thank you for this invaluable contribution :) Some of our Indigenous community users will be very excited to see it.

rudokemper:
I have tested out this PR and can confirm that this achieves the desired behavior. I see the error being thrown by removing a Photo that @EitanFuturo describes as well; and, that the overall app UI when it comes to removing/replacing existing media content is not the most intuitive:

image

In the future, for these fields that only accept one file, we may want to disable the `Choose File` button until the user has removed any existing media content. (After the above error is fixed across the board.)

And definitely some more food for thought here about continued use of Administrate that we've discussed with @FeminismIsAwesome @lauramosher.

All that aside, I think this is a great PR and just flagging this for additional code review.

rudokemper (approved):
LGTM!

PR #713, opened 2022-03-03
Fix issue with local IP in development environment

andersan:
No impact on production environment. Local IP was not recognized, and delete actions in administrate views were not available. This commit fixes that.


andersan:
Added a second commit which fixes attachment removal

lauramosher (approved):
Thank you for this!!

PR #716, opened 2022-04-07
Update dockerfile

lauramosher:
For the most part, this intentionally changes very little.

Most notable change is that the Dockerfile is entirely different:

1. It uses a base image of Alpine Linux instead of debian based ubuntu (or similar)
2. Bumps Ruby 2 minor versions (2.5.1 -> 2.5.3) in order to take advantage of the built in node and yarn packages (with the correct versions). This isn't the latest minor, but I wanted to avoid bumping anything too far without more testing.

Because Alpine doesn't come with bash by default, the script files have all been changed to /bin/sh instead. The line that changed mod for MacOS didn't seem to be need anymore, so I removed that from this Dockerfile.

I was able to fully test this in macOS Monterey intel from nothing to seeded and running. I have an M1 Mac we can test on, as well, but wanted to do all of the updates using intel/amd based first.

@rudokemper Can you test this on Windows? I no longer have a windows machine to test on.


rudokemper:
Thanks @lauramosher will take a look tomorrow!

rudokemper (approved):
Works great on Windows! LGTM

PR #718, opened 2022-04-08
Update ruby and ruby on rails

lauramosher:
This is bare minimum upgrades:

- First ensured we were on the latest Rails 5.2.X version.
- Resolved any remaining deprecation warnings.
- Upgrade ruby version one at a time, even through minor versions, just to be careful.
- Upgraded to Rails 6.0.0
- Ran through upgrade list, per [Rails upgrade guide](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-2-to-rails-6-0). Remained on 5.2 loading defaults and classic mode.
- Once things worked, flipped 6.0 defaults; particularly ones we don't use in this app
- Upgraded to latest 6.0.X version patch
- Enabled zeitwerk and ran check. Once all good, committed to load 6.0 defaults which uses zeitwerk by default.
- Finally fixed import Docker volume that was broke in #716


rudokemper (approved):
LGTM

Issue #618, opened 2021-04-14
[Map] Mapbox GL JS v2 / 3D Terrain integration in Terrastories


rudokemper:
Let's add support for using the amazing new Mapbox GL JS v2 engine in Terrastories, so we can leverage the 3D terrain functionality for online Terrastories builds!

This requires scoping out to see what the best approach is -- an environmental parameter to enable 3D terrain, or something else. But logging this as an issue for now.

PR #770, opened 2022-04-14
Add 3D Terrain toggle for online maps

rudokemper:
Closes https://github.com/Terrastories/terrastories/issues/618.

This PR adds a new feature to turn 3d Terrain for Mapbox.com maps on and off. Mapbox's 3D terrain is a new, realistic 3D exaggeration layer based on a digital elevation model, and adds another level of detail to the landscape and where stories may be found. It can however be a performance hog, so it will be best to allow our users to decide whether they want this feature activated or not. This PR makes it possible to do so for an admin user, by creating a boolean in the Theme model and menu. 

* `mapbox_3d` boolean added to Theme model.
* On the React side, if `mapbox_3d` is activated, then the map will render in 3d.

In the future we could explore having this be a setting per user instead of per community, but I think it would be good to get some user input on this decision and the feature as a whole.

image

3d Terrain activated:
image

3d Terrain de-activated:
image

### DIFF
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -23,6 +23,7 @@ export default class Map extends Component {
     onMapPointClick: PropTypes.func,
     mapboxStyle: PropTypes.string,
     mapboxAccessToken: PropTypes.string,
+    mapbox3d: PropTypes.bool,
     useLocalMapServer: PropTypes.bool,
     markerImgUrl: PropTypes.string,
   };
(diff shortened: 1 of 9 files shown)


rails/app/javascript/components/Map.jsx
```diff
@@ -57,6 +58,28 @@ export default class Map extends Component {
           "icon-allow-overlap": true
         }
       });
+      if(this.props.mapbox3d) {
+        this.map.addSource('mapbox-dem', {
+          'type': 'raster-dem',
+          'url': 'mapbox://mapbox.mapbox-terrain-dem-v1',
```
lauramosher, line 64: I saw the note that this only works for online maps. Should we be concerned about this source / layer attempting to be added when offline?

lauramosher (approved):
Just one comment about offline mode with the mapbox url. Otherwise, this looks great! 👍

Issue #703, opened 2021-11-03
[Rails][Mapbox] Permit null values for the map bounding box in a Community theme


rudokemper:
Currently, we have some validations set up that forces a numerical value for the bounding box attributes in Theme. These are `sw_boundary_lat`, `sw_boundary_long`, `ne_boundary_lat`, and `nw_boundary_long`. These feed into a bounding box that delimits the maximum limits of where you can pan on the map.

This is a useful thing to have as an option, but we should start with no bounding box to begin with so that the user can pan infinitely across the globe (like so: https://share.getcloudapp.com/NQu7w298).

To make this happen, we first need to allow the user to set null values for these fields in the validations. 
That should allow the map to function as in the example above. 
Lastly, we should have null default bounding box values for a new community (as set in `map_configurable.rb`)..

## Context:
Currently, with the standard map default bounding box values set in `map_configurable.rb` ((-85, -180), (85, 180)), we are boxing communities into a rectangular view of the map, which cuts the Pacific ocean in half.

## Acceptance criteria:

- [ ] If there are no values set for the four bounding box attributes, the map should permit infinite horizontal panning across the globe, and not cut off at the Pacific ocean.
- [ ] In the Theme dashboard, a user is able to set _either_ values within the accepted range for coordinates (as set currently in `theme_spec.rb`), or null.
- [ ] When a new community is created, the default bounding box values are null.

willeyen:
I would like to work on this issue, if possible!

PR #772, opened 2022-04-19
IN-PROGRESS (feedback requested) Issue #703 Fix - Default Null Bounding Boxes for Map

willeyen:
This is a work-in-progress; I am hoping for feedback and review on my proposed changes. Specifically, I am unsure of the validation scheme I implemented in theme.rb (see second bullet point below).

Summary: changed default bounding box coordinates to be null for a new community, allowed for bounding boxes to be updated in theme to and from null (theme.rb), changed map construction to account for possibility of no/null bounding box (Map.jsx), changed theme_spec.rb to account for null default boundaries.

Changes:

- First, I changed the map_configurable.rb file such that the default values for sw_boundary_lat, sw_boundary_long, ne_boundary_lat, and ne_boundary_long are all now nil. This allows for infinite horizontal panning across the globe as desired. I also checked the functionality when creating a new community, and it does set the bounding box coordinates to null when a new community is created, as desired. I made some other changes in this file so that the boundaries should only be valid/enforced when all four coordinates are set. 

- I’m not sure about the validation I set up in theme.rb. The goal was to only validate the new boundaries if either 1) the 4 boundaries were all set to null OR 2) the 4 boundaries were all numeric and in the proper ranges. However, when I changed allow_nil to false in the numeric case, this caused the 1st validation to always fail — making it impossible to go back to having no bounding boxes / null coordinates. So, for now I kept allow_nil as true in the numeric case, which means that users can set some of the conditions to numbers and some of the conditions to null without seeing an error message. I think this just keeps the previous proper validation of the map in place until a new valid set of boundaries is provided, and map functionality is not disrupted to my current knowledge. However, the theme editing page displays the incorrectly set values and an error message is not issued. So I’m not sure if this is truly the desired functionality or how to fix this, as I tried a few different validations that performed worse than this (i.e., caused the map to malfunction and in some cases allowed out-of-range numbers) without solving the issue. I’m also new to validations/Ruby so there may something obvious I’m missing! I would love feedback on this point. 

- I’m also unsure of the changes I made to theme_spec.rb. I’m not sure when this file is triggered/called, but it seemed reasonable to change the expected values to nil for the sw_boundary and ne_boundary since nil is the new default for the boundaries/bounding box coordinates set in map_configurable.rb.

- In Map.jsx, I changed map construction in the componentDidMount() function to account for the possibility of no/null bounding box coordinates. I did the same for the resetMapToCenter() function, as both of these functions set maxBounds based on the 4 boundary coordinates. #703

### DIFF
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -33,7 +33,7 @@ export default class Map extends Component {
         style: this.props.mapboxStyle,
         center: [this.props.centerLong, this.props.centerLat],
         zoom: this.props.zoom,
-        maxBounds: this.checkBB(), // check for bounding box presence
+        maxBounds: this.checkBounds(), // check for bounding box presence
         pitch: this.props.pitch,
         bearing: this.props.bearing
     });
@@ -154,7 +154,7 @@ export default class Map extends Component {
         zoom: this.props.zoom,
         pitch: this.props.pitch,
         bearing: this.props.bearing,
-        maxBounds: this.checkBB(), // check for bounding box presence
+        maxBounds: this.checkBounds(), // check for bounding box presence
     });
   }
 
@@ -191,7 +191,7 @@ export default class Map extends Component {
   }
 
 // test for bounding box presence
-  checkBB() {
+  checkBounds() {
     let mapBounds = null;
     if (this.props.sw_boundary_long != null && this.props.sw_boundary_lat != null
         && this.props.ne_boundary_long != null && this.props.ne_boundary_lat != null) {
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -28,30 +28,15 @@ export default class Map extends Component {
   };
 
   componentDidMount() {
-    if (this.props.sw_boundary_long == null) { // no bounding box
-      this.map = new mapboxgl.Map({
+    this.map = new mapboxgl.Map({
         container: this.mapContainer,
         style: this.props.mapboxStyle,
         center: [this.props.centerLong, this.props.centerLat],
         zoom: this.props.zoom,
-        maxBounds: null, // no bounding box
+        maxBounds: this.checkBB(), // check for bounding box presence
         pitch: this.props.pitch,
         bearing: this.props.bearing
-      });
-    } else { // map does have bounding box
-      this.map = new mapboxgl.Map({
-        container: this.mapContainer,
-        style: this.props.mapboxStyle,
-        center: [this.props.centerLong, this.props.centerLat],
-        zoom: this.props.zoom,
-        maxBounds: [
-          [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
-          [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
-        ],
-        pitch: this.props.pitch,
-        bearing: this.props.bearing
-      });
-    }
+    });
 
     this.map.on("load", () => {
 
@@ -164,26 +149,13 @@ export default class Map extends Component {
   }
 
   resetMapToCenter() {
-    if (this.props.sw_boundary_long == null) { // no bounding box
-      this.map.flyTo({
+    this.map.flyTo({
         center: [this.props.centerLong, this.props.centerLat],
         zoom: this.props.zoom,
         pitch: this.props.pitch,
         bearing: this.props.bearing,
-        maxBounds: null // no bounding box
-      });
-    } else { // map does have bounding box
-      this.map.flyTo({
-        center: [this.props.centerLong, this.props.centerLat],
-        zoom: this.props.zoom,
-        pitch: this.props.pitch,
-        bearing: this.props.bearing,
-        maxBounds: [
-          [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
-          [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
-        ]
-      });
-    }
+        maxBounds: this.checkBB(), // check for bounding box presence
+    });
   }
 
   // TODO: update this to JSX
@@ -216,5 +188,18 @@ export default class Map extends Component {
 
   render() {
     return <div ref={el => (this.mapContainer = el)} className="ts-MainMap" />;
+  }
+
+// test for bounding box presence
+  checkBB() {
+    let mapBounds = null;
+    if (this.props.sw_boundary_long != null && this.props.sw_boundary_lat != null
+        && this.props.ne_boundary_long != null && this.props.ne_boundary_lat != null) {
+        mapBounds = [
+            [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
+            [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
+        ]
+    }
+    return mapBounds;
   }
 }
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -28,18 +28,30 @@ export default class Map extends Component {
   };
 
   componentDidMount() {
-    this.map = new mapboxgl.Map({
-      container: this.mapContainer,
-      style: this.props.mapboxStyle,
-      center: [this.props.centerLong, this.props.centerLat],
-      zoom: this.props.zoom,
-      maxBounds: [
-        [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
-        [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
-      ],
-      pitch: this.props.pitch,
-      bearing: this.props.bearing
-    });
+    if (this.props.sw_boundary_long == null) { // no bounding box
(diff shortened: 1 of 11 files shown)


rails/app/javascript/components/Map.jsx
```diff
@@ -28,18 +28,30 @@ export default class Map extends Component {
...
+        center: [this.props.centerLong, this.props.centerLat],
+        zoom: this.props.zoom,
+        maxBounds: [
+          [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
+          [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
+        ],
+        pitch: this.props.pitch,
+        bearing: this.props.bearing
+      });
+    }
```
rudokemper, line 54: I might suggest simplifying this code and `resetMapToCenter()` somewhat. Instead of repeating the map definition, you could have some code before like:

```
    let mapBounds = null; 
    if (this.props.sw_boundary_long != null) {
      mapBounds = [
      [this.props.sw_boundary_long, this.props.sw_boundary_lat], //southwest
      [this.props.ne_boundary_long, this.props.ne_boundary_lat] //northeast
      ]
    }
```

And then set `maxBounds` to `mapBounds`.

Or maybe using that `boundary` definition?

Also, here we might want to check for all of the boundary lat/longs, not just `sw_boundary_long`.

rudokemper:
This is looking great @willeyen! I tried it out and it is working as desired. I left a comment on the `Map.jsx` file targeting simplifying the code a bit. @lauramosher might you be able to take a look at the Ruby validations?

rails/app/javascript/components/Map.jsx
willeyen: I just pushed a new commit simplifying the code and checking for all of the boundary lat/longs! Thanks for the feedback.

rails/app/models/theme.rb
```diff
@@ -9,9 +9,18 @@ class Theme < ApplicationRecord
...
-  validates :center_lat, :sw_boundary_lat, :ne_boundary_lat,
+  if :sw_boundary_long == nil and :sw_boundary_lat == nil and :ne_boundary_lat == nil and :ne_boundary_long == nil
+    validates :sw_boundary_long, :ne_boundary_long, :sw_boundary_lat, :ne_boundary_lat,
+      absence: true # update to remove bounding box if all bb coordinates are empty, else enforce constraints
+  else
+    validates :sw_boundary_lat, :ne_boundary_lat,
+        :numericality=> true, allow_nil: true, :inclusion => {:in => -90..90, :message => "value should be between -90 and 90"}
+    validates :sw_boundary_long, :ne_boundary_long,
+        :numericality=> true, allow_nil: true, :inclusion => {:in => -180..180, :message => "value should be between -180 and 180"}
+  end
```
viamin, line 20: I don't think this will work, since (I believe) the `if/else` statements will be evaluated when the model is loaded (that is, on app initialization) rather than when the model is validated. That means the `else` branch will always be active (since `:sw_boundary_long == nil` will _always_ be false) 

However you can use [conditional validation](https://guides.rubyonrails.org/active_record_validations.html#conditional-validation). Alternately you could write a [custom validator class](https://guides.rubyonrails.org/active_record_validations.html#performing-custom-validations) that checks everything, but given what you've already written the first way is probably easier:
```suggestion
  validates :sw_boundary_long, :ne_boundary_long, :sw_boundary_lat, :ne_boundary_lat,
    absence: true,
    if: :nil_boundaries?
    # update to remove bounding box if all bb coordinates are empty, else enforce constraints
  validates :sw_boundary_lat, :ne_boundary_lat,
      numericality: true,
      allow_nil: true,
      inclusion: {in: -90..90, message: "value should be between -90 and 90"},
      unless: :nil_boundaries?
  validates :sw_boundary_long, :ne_boundary_long,
      numericality: true,
      allow_nil: true,
      inclusion: {in: -180..180, message: "value should be between -180 and 180"},
      unless: :nil_boundaries?
      
    def nil_boundaries?
      sw_boundary_long.nil? and sw_boundary_lat.nil? and ne_boundary_lat.nil? and ne_boundary_long.nil?
    end
```

viamin: Thinking a little bit more about this, I think you can get away with only the last 2 validations with the conditional, and change `allow_nil` to `false`. This should satisfy what you mentioned in bullet 2 (either all are nil, or none are and are numerical within the proper range - am I reading that right?) The first validation would just be redundant (validate all 4 values are nil if all four values are already nil.)
```suggestion
  validates :sw_boundary_lat, :ne_boundary_lat,
    numericality: true,
    allow_nil: false,
    inclusion: {in: -90..90, message: "value should be between -90 and 90"},
    unless: :nil_boundaries?
  validates :sw_boundary_long, :ne_boundary_long,
    numericality: true,
    allow_nil: false,
    inclusion: {in: -180..180, message: "value should be between -180 and 180"},
    unless: :nil_boundaries?
    
  def nil_boundaries?
    sw_boundary_long.nil? and sw_boundary_lat.nil? and ne_boundary_lat.nil? and ne_boundary_long.nil?
  end
```

lauramosher: > I don't think this will work, since (I believe) the if/else statements will be evaluated when the model is loaded (that is, on app initialization) rather than when the model is validated. That means the else branch will always be active (since :sw_boundary_long == nil will always be false)

I believe even if the entire conditional was moved into a conditional validation and loaded at call, it would still always be false. Symbols defined as they are here are values *as themselves* and so would never be nil.

e.g.

```ruby
:test
=> :test
:test.nil?
=> false
```

I really like the adjusted validations that @viamin suggested with only using the two validations and disallowing nil unless all 4 values are nil.

We could also take it a step further and handle this in a [custom validation method](https://guides.rubyonrails.org/v6.0/active_record_validations.html#custom-methods) instead (not to be confused with a custom validator class). This would make it extremely obvious that the map bounds are all required, or all nil. The fact that we have two separate validations that then use the same `nil_boundaries?` check feels like this may be a warranted move.

Something akin to:

```suggestion
  validate :map_bounds
  
  def map_bounds
    return if sw_boundary_long.nil? && sw_boundary_lat.nil? # southwest bounds
       && ne_boundary_long.nil? && ne_boundary_lat.nil? # northeast bounds
    
    errors.add(:sw_boundary_lat, "value should be between -90 and 90") unless (-90..90).include?(sw_boundary_lat)
    errors.add(:ne_boundary_lat, "value should be between -90 and 90") unless (-90..90).include?(ne_boundary_lat)
    errors.add(:sw_boundary_long, "value should be between -180 and 180") unless (-180..180).include?(sw_boundary_long)
    errors.add(:ne_boundary_long, "value should be between -180 and 180") unless (-180..180).include?(ne_boundary_long)
  end
```

lauramosher (changes requested):
Thank you so much for your contribution! This change will be so helpful for our communities.

> I’m also unsure of the changes I made to theme_spec.rb. I’m not sure when this file is triggered/called, but it seemed reasonable to change the expected values to nil for the sw_boundary and ne_boundary since nil is the new default for the boundaries/bounding box coordinates set in map_configurable.rb.

The file is only called/evaluated when running our automated test suite... and since our tests are not currently fully functional or fully passing at the moment, I would agree that this change is reasonable for the scope of this PR.

Per your question re: validations, I've added additional comments inline for working with validations and achieving the all or nothing map bounds.

Let us know if you have any questions.

willeyen:
Thank you @viamin and @lauramosher for your help! I changed the bounding box validation to use the custom validation method as suggested. It works locally for me now!

rails/app/models/theme.rb
```diff
@@ -9,9 +9,19 @@ class Theme < ApplicationRecord
   validates :sponsor_logos, blob: { content_type: ['image/png', 'image/jpg', 'image/jpeg'], size_range: 1..5.megabytes }
   validates :mapbox_access_token, :presence =>  {:message => 'is required when the Mapbox style URL is set.'}, unless: -> { mapbox_style_url.blank? }
   validates :mapbox_style_url, :presence => {:message => 'is required when the Mapbox access token is set.'}, unless: -> { mapbox_access_token.blank? }
-  validates :center_lat, :sw_boundary_lat, :ne_boundary_lat,
+  validate :map_bounds
+   # validate bounding box if all four values are nil OR if all four values are numeric & in the proper range
+   def map_bounds
+     return if sw_boundary_long.nil? && sw_boundary_lat.nil? && ne_boundary_long.nil? && ne_boundary_lat.nil?
+
```
rudokemper, line 16: I would add one more validation here:

`errors.add(:base, "All four bounding box values must be set, or left blank")`

rudokemper (changes requested):
Hi @willeyen! I tested this out locally and it works just as expected. Thanks so much for your work on this. I just added one more validation suggestion (to cue the user that leaving the bound box fields blank is also an option) and it should be good to go.

rails/app/models/theme.rb
willeyen: I just committed a change to add this validation! The message will show if the user submits a mixture of nil values and numeric values.

rudokemper (approved):

lauramosher (approved):
This looks great! Thank you so much!

PR #778, opened 2022-05-01
In-Progress: Issue #579 - Cluster Overlapping Points on Map

willeyen:
I added code for clustering map points in Map.jsx, using some of the code from here: https://docs.mapbox.com/mapbox-gl-js/example/cluster/

As requested, clicking on a cluster zooms in on it.

For sizing, I used the interpolation function to scale up the size of each cluster marker to indicate the number of points within it. The scaling values I chose (somewhat arbitrarily, but I tested a few and these looked best for me) worked locally for me, but the sizing might be off for some communities. The sizing could also just be set to be uniformly larger for any size cluster than the single marker.

Also, the cluster radius is set to the default 50 right now. I think changing this value would cause more or less clusters to form. I'm not sure if there's a way to set it so that clusters form only if the individual story markers are actually overlapping, but the default value works well enough for me locally. Again, with an actual community this value may not be best. (I tested locally with 22 points in Western Kentucky, so a medium-sized area.)

I'll be busy the next few weeks with finals, graduation, and moving to a new city, so feel free to pass this issue along to someone else to finish it! I'm also happy to try to implement feedback when I have the time, whatever works best for you all :)


rudokemper:
Hi @willeyen, this is great! Thanks so much for this work. I tested your branch, and as I was playing around with things, I ended up making a few changes that I think can be added to your PR: adding a cluster icon and a count of points to cue the user and differentiate between the story points.

image

In the future, we will need to think about how to visually "stack" a user's own custom icons when we make it possible for users to do that. But for now, this cluster icon PNG can work for an MVP for clustering.

I will file a PR on your branch 😃

rudokemper (changes requested):
Once the open PR on this branch is merged and any conflicts resolved, I think this is good to go. I'll do one more round of testing on it with some additional datasets/maps, and if that looks good, will approve and merge.

rudokemper (approved):

PR #791, opened 2022-08-13
790: Add map projection to Rails and front end (WIP)

### LINKED ISSUE #790: [Map] Add ability to set map projection (and prevent new globe projection from temporarily breaking)
## Context:
Starting from v2.6, Mapbox GL JS supports multiple map projections. This feature allows you to create more accurate visualizations at every zoom level and tell a better story with your data.

We would like to be able to permit a Terrastories community admin user to set the projection for their map.

In addition, the brand new `globe` projection is currently breaking Terrastories. Mapbox introduced this new projection in Mapbox GL JS 2.9, and new maps in Mapbox Studio are created with this projection included by default. Hence, unless we override this projection in Terrastories, these new maps will not load until `mapbox-gl-rails` updates to 2.9. So adding a feature for a Terrastories admin user to set the projection (with a default of `mercator` and a select menu excluding `globe` for the time being) solves this problem as well.

## Acceptance criteria:
- [ ] A new projection field is added to `themes`
- [ ] When a new community is created, the default projection is set to `mercator`
- [ ] An admin user for a community is able to change the projection to the other projections supported as of Mapbox GL JS 2.7 in `Themes` menu.
- [ ] When loading the map, it will be projected using whichever projection is selected by the admin user.
- [ ] Ensure that the projection change works offline with TileServer-GL.

rudokemper:
Closes #790.

This PR adds a field `map_projection` to the `themes` model, sets it to `mercator` by default, and permits an admin user to change the projection in the Themes menu to one of six possible projections that are supported by Mapbox GL JS 2.7. The map on the front end will then use that projection set by the admin user.

image

Natural Earth:
image

Albers:
image

Lambert Conformal Conic:
image

Works with default offline vector tiles and TileServer-GL while entirely offline:
image

Once `mapbox-gl-rails` updates to 2.9, we can file a small PR to add the `globe` projection as well, but will likely have to account for this not working with TileServer-GL when offline.

Lastly, a few small housekeeping fixes:
* Move the `zoom`, `pitch`, and `bearing` fields above the bounding box fields in the Theme menu, since these are more useful to the user than the bounding box.
* Fixing broken links in documentation
* Reset 3d terrain exaggeration to 1 (default)


rudokemper:
Will update this PR to build on the new refactor code, next week.

PR #816, opened 2022-10-03
Fix: Error with Importable header mismatch rescue

lauramosher:
Fixes error on Heroku boot:
```
app/controllers/dashboard/imports_controller.rb:3:in <class:ImportsController>': uninitialized constant Dashboard::ImportsController::FileImporter (NameError)
```


rudokemper:
Tested this out on Heroku and LGTM ✅

PR #818, opened 2022-10-03
Fix: Allow lat/long values to accept any valid latitude or longitude value

lauramosher:
This ensures the input field for lat and long can accept any valid latitude or longitude value.

Before, the input would through a client-side validation error if the value was not a whole number. See discussion below for details.

### DIFF
--- a/rails/app/views/dashboard/places/_form.html.erb
+++ b/rails/app/views/dashboard/places/_form.html.erb
@@ -36,10 +36,10 @@
   </div>
 
   <%= f.label :lat %>
-  <%= f.number_field :lat, step: "0.00001", min: -90, max: 90 %>
+  <%= f.text_field :lat %>
 
   <%= f.label :long %>
-  <%= f.number_field :long, step: "0.00001", min: -180, max: 180 %>
+  <%= f.text_field :long %>
 
   <%= f.submit %>
 <% end %>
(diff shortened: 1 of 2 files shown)


rudokemper:
Thanks @lauramosher!

Is there a way that the fields can accept values with more than 5 decimal places, but either round it up to 5 when saving, or, only show 5 decimal places to the user? 

I think I might have miscommunicated on this; it's definitely not necessary to show more than 5 decimal fields (e.g. for the coords selected by the map in Themes), but if a user has a lat/long value with >5 decimal places, it's still a valid coordinate. Right now this validation is rejecting it.

(User story here is that >5 decimal places for coordinates become more and more trivial for the purposes of Terrastories - once you get to the 6th decimal point, it's so precise that it's superfluous. Like the level of standing in one side of the house vs. the other 😆)

lauramosher:
Ah! Sure. We can just make them text fields then! The number field has built in client-side validation which is what we were running into.

lauramosher:
@rudokemper I've updated this PR to use text fields instead of number fields. It also updates the theme form as well.

Let me know if this all works. I'll test it soon (rebuilding my local instances) myself, as well.

rudokemper (approved):
Tested lat/long in both `Places` and `Themes` and this is working as desired! 🎉

PR #830, opened 2022-10-04
fix: zoom issue with map

asheerrizvi:


rudokemper (approved):
Awesome, this solves the issue; thanks @asheerrizvi!

PR #831, opened 2022-10-04
Fix: Stories Filtering

lauramosher:
Ensure that StoryPage correctly filters habtm relationships so ALL relations are shown even if only matched on one.

Essentially, rather than filtering from Story -> PlaceStory -> Place, we only go Story -> PlaceStory. Same for Speakers.

| Before | After |
| --- | --- |
| image | image |


rudokemper (approved):
LGTM!

PR #832, opened 2022-10-04
Feat: Mark required fields

lauramosher:
This marks form fields as required with a red asterisk:

image


rudokemper (approved):
Just test drove - LGTM

PR #833, opened 2022-10-04
Feat: add map projection settings to Terrastories

rudokemper:
Closes https://github.com/Terrastories/terrastories/issues/790.

This PR adds a field `map_projection` to the `themes` model, sets it to `mercator` by default, and permits an admin user to change the projection in the Themes menu to one of six possible projections that are supported by Mapbox GL JS 2.7. The map on the front end will then use that projection set by the admin user.

image
Natural Earth:
image

Albers:
image

Lambert Conformal Conic:
image

Works with default offline vector tiles and TileServer-GL while entirely offline:
image

Once `mapbox-gl-rails` updates to 2.9, we can file a small PR to add the `globe` projection as well, but will likely have to account for this not working with TileServer-GL when offline.

Lastly, a few small housekeeping fixes:
* Move the `zoom`, `pitch`, and `bearing` fields above the bounding box fields in the Theme menu, since these are more useful to the user than the bounding box.
* Fixing broken links in documentation
* Reset 3d terrain exaggeration to 1 (default)

### DIFF
--- a/rails/config/locales/en/administrate.en.yml
+++ b/rails/config/locales/en/administrate.en.yml
@@ -34,6 +34,7 @@ en:
         mapbox_style_url: Mapbox style URL (for online maps)
         mapbox_access_token: Mapbox access token associated with the style
         mapbox_3d: Activate 3D Terrain view for online map
+        map_projection: Set projection for map
         center_lat: Map center, latitude
         center_long: Map center, longitude
         sw_boundary_lat: SW bounding box, latitude (optional)
--- a/rails/config/locales/en/dashboard.en.yml
+++ b/rails/config/locales/en/dashboard.en.yml
@@ -12,13 +12,14 @@ en:
     online: Online Map Configuration
     settings: Map Settings
     settings_description: Customize your community's map settings.
-    mercator: Mercator
-    albers: Albers
-    equalEarth: Equal Earth
-    equirectangular: Equirectangular
-    lambertConformalConic: Lambert Comformal Conic
-    naturalEarth: Natural Earth
-    winkelTripel: Winkel Tripel
+    projection:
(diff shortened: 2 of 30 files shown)


rails/config/locales/en/administrate.en.yml
```diff
@@ -34,6 +34,7 @@ en:
         mapbox_style_url: Mapbox style URL (for online maps)
         mapbox_access_token: Mapbox access token associated with the style
         mapbox_3d: Activate 3D Terrain view for online map
+        map_projection: Set projection for map
```
lauramosher, line 37: Let's populate this translation in `models.en.yml` as well. 

We should also probably add the key to the other language files as well even if the translated value is currently English.

rails/config/locales/en/dashboard.en.yml
```diff
@@ -12,6 +12,14 @@ en:
...
     settings: Map Settings
     settings_description: Customize your community's map settings.
+    projection:
+      mercator: Mercator
+      albers: Albers
+      equalEarth: Equal Earth
+      equirectangular: Equirectangular
+      lambertConformalConic: Lambert Comformal Conic
+      naturalEarth: Natural Earth
+      winkelTripel: Winkel Tripel
```
lauramosher, line 22: Let's add these to all of the other translation files, too, other we'll get translation missing texts.

lauramosher (changes requested):

rudokemper:
@lauramosher oops, good catch.
I saw there was a bunch of Theme translation stuff missing for administrate in MAT, NL, JA, and added those too. Not that it will matter for long 😛

lauramosher (approved):
👍

PR #835, opened 2022-10-04
Feat: allow signin with username (in addition to email)

### LINKED ISSUE #793: [Rails] Let's create a string for Terrastories users called username and use that for logging in/out instead of email
### Discussed in https://github.com/Terrastories/terrastories/discussions/759

<div type='discussions-op-text'>

<sup>Originally posted by **rudokemper** February  6, 2021</sup>
## Context:
Currently, Terrastories users are being asked to log in or register with an email address.  

Most of our current users (i.e. communities in the Amazon) don't have an email address, and so this is not an appropriate format for their use case. In addition, some of our potential users inquiring about Terrastories have concerns about security and would rather not submit an email address. 

Let's make a string `username` in the `users` table and make that be the field that people use to log in and out to Terrastories, instead of `email`.</div>

### LINKED ISSUE #797: Sign in with username in addition to email
Allow a user to sign in with a username in addition to email.

lauramosher:
Resolves #797
Resolves #793

This adds the ability for a user to sign in with a username instead of an email. Logging in with email still works!

For backwards compatibility, this saves the user's current email to the username attribute. The migration is reversible, even after adding new users with usernames without email addresses.

Additionally, this:

1. Fixes an issue with images not displaying correctly across the dashboard (found during user updates)
2. Adds a new /profile route for the current user to manage their own user profile without needing to display user ID in the route; only available in the dropdown in email
3. Adds username AND name field to User. App will display name when available and fallback to username.
4. Fixes a flash message error where things would be unclickable underneath an invisible element.5. 
5. Updates Devise registration new and edit forms to allow sign up and account update with new attributes.6. 
6. Updates Welcome / No-Community user to add a profile link to update their profile information.

https://user-images.githubusercontent.com/2660801/193943464-b9e7422c-e433-4180-82ca-afbf5e9819a8.mp4

### DIFF
--- a/rails/app/assets/stylesheets/cms/_flash.scss
+++ b/rails/app/assets/stylesheets/cms/_flash.scss
@@ -3,7 +3,6 @@ div.flash-messages {
   width: 100%;
   display: grid;
   place-items: center;
-  z-index: 1000;
 }
 
 span.flash {
@@ -14,6 +13,7 @@ span.flash {
   margin: 1rem;
   border: 1px solid $dark-green;
   opacity: 1;
+  z-index: 1000;
 
   max-width: 30vw;
 
--- a/rails/app/controllers/dashboard/users_controller.rb
+++ b/rails/app/controllers/dashboard/users_controller.rb
@@ -76,6 +76,8 @@ module Dashboard
 
     def user_params
       params.require(:user).permit(
+        :name,
+        :username,
         :email,
         :password,
         :role,
--- a/rails/app/controllers/dashboard/users_controller.rb
+++ b/rails/app/controllers/dashboard/users_controller.rb
@@ -41,6 +41,11 @@ module Dashboard
       @user = authorize community.users.find(params[:id])
     end
 
+    def profile
+      @user = current_user
+      render :edit
+    end
+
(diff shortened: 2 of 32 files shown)


rudokemper (changes requested):
This looks great! So excited to finally have this in place.

One small fix - since we are now requiring `username` for users, we need to add usernames to the seed data as well. I built this branch with a blank database and it would not build until I added usernames for all of the seeded users.

Let's do `terrastories-super`, `terrastories-admin`, `terrastories-editor`, `terrastories-user`, `terrastories-viewer`, and `terrastories-public` for the seeds, respectively.

Backwards compatibility handling looks good
~~, however it appears there is a bug here - I see that for one existing community on staging towards which the backwards compat migration was applied, there is some weird behavior where one of the usernames inherits the wrong email and password. (See screenshot below). Interestingly, for a split second, it shows the right email address and password, but then it switches to the wrong ones. Same thing on Administrate. Will reach out with the login to check this out.~~
 NVM just some annoying behavior from browser LastPass plugin!

rudokemper (approved):
Great - LGTM!

PR #836, opened 2022-10-05
Truncates speaker and place name to 30 characters in the story filters

### LINKED ISSUE #822: [Rails] Truncate Story filter select options in new CMS
The new Terrastories CMS (accessed at `/member/`) adds a new filter tool for filtering through Stories.

Currently, the three select boxes for this filter tool for Place and Speaker are populated by the names from records in these two models, and depending on the length of the names, the select boxes can grow quite large, and could result in undesired overflow of the filter tool past the screen width.

Let's truncate the length of the names that are shown in these boxes to 30 characters.

image

spencerldixon:
Resolves #822 

Truncates the speaker and place names in the story filters to a maximum of 30 characters to prevent filter fields getting too long.

### DIFF
--- a/rails/app/views/dashboard/stories/index.html.erb
+++ b/rails/app/views/dashboard/stories/index.html.erb
@@ -10,11 +10,11 @@
   <%= form_with url: stories_path, method: :get, class: "filters", local: true do |f| %>
     <div>
     <%= f.label :place, t("place") %>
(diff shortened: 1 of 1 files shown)


spencerldixon:
Hey @lauramosher thanks for the feedback! That makes a lot more sense, I've reverted those changes and added a max-width to the filters with css. They appear at a nice size now, but still display the full length of the name when the dropdown is expanded

Issue #826, opened 2022-10-04
[Translation] Help us add more translations to Terrastories! (and update Spanish/Portuguese/Dutch/Japanese)


rudokemper:
We've been slacking on keeping our Spanish, Portuguese, Dutch, and Japanese translations in `/rails/config/locales` up to date as we've been adding more strings to the app. Also, we would love to add more languages to Terrastories! If you would either like to help us keep our translations complete, or add a new language, that would be amazing!

Process for adding a new language:
1. Find the two or three-digit ISO language code for your language contribution [here](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
2. Copy the `en` folder and rename it to your ISO language code. Also rename the `en` in each filename, and the beginning of each file, to your ISO language code.
3. Translate each string to your language of contribution.
4. Add your language to the `languages` list for all of the languages (in `en.yml`, `pt.yml`, `mat.yml`, ... and your new language contribution). You can use Google Translate to figure out how your language is translated in the other languages.
5. Submit your PR 🎉 

Please note that we can't merge PRs until at least the core content of `en.yml` has been translated, since that is what feeds the front end of the app. It is recommended to submit a PR with all of the translations at once, but if you want to do them piece-meal instead, please start with `en.yml`.

Updates:
- [X] Spanish
- [X] Dutch
- [X] Matawai
- [X] Chinese
- [x] Portuguese
- [ ] Japanese

larymak:
Can I add a Swahili translation to the same?

rudokemper:
@larymak absolutely! We would love to have Swahili!

KrunkZhou:
Can I add a Chinese translation?

rudokemper:
@KrunkZhou that would be wonderful! :)

julinvictus:
I'm doing the PT one

julinvictus:
Wanted to check with you @rudokemper what you think about trying to use gender-neutral expressions. Latin languages like Portuguese will always genderize everything, but there's a tendency now to try to use those gender-neutral expressions, for instance, Welcome screen could be 'Bem vinde', now is 'Bem vindo/a'.

rudokemper:
Great question @julinvictus. I thought about this, and went back and forth on it a little. On the one hand, some of our users are not likely to know gender-neutral expressions, but on the other hand, we want to be inclusive and I checked with some peer applications and platforms that are already using these as well. So I'd say let's go with the gender neutral expressions 👍 thanks for asking!

[...]

PR #838, opened 2022-10-05
feat:(translation) ES - Update translation for file es.yml

hereje:
Update spanish translation for file

`rails/config/locales/es/es.yml`

#826


lauramosher (approved):
Thank you!

PR #843, opened 2022-10-07
fix: format and add `username` field to create community form

rudokemper:
This PR formats, and adds a `username` field to the super-admin's create community form.

Without the `username` field (as a now required field for users), super admin is not able to create any communities as it flashes an error upon saving `Users username can't be blank`.

image


rudokemper:
@lauramosher I have no doubt that you will know right away why `:username` field is not being saved in this form despite being included.

lauramosher:
Yes, you’ll need to add username to the strong parameters in the controller.

lauramosher (approved):

PR #846, opened 2022-10-07
fix: show display name (if not username) instead of email on front end

rudokemper:
Show the logged in user's display name (if not username) instead of email on the front end.

image

### DIFF
--- a/rails/app/javascript/components/App.jsx
+++ b/rails/app/javascript/components/App.jsx
@@ -379,6 +379,7 @@ class App extends Component {
           onStoryClick={this.handleStoryClick}
           logo_path={this.props.logo_path}
           user={this.props.user}
+          displayName={this.props.display_name}
           filterCategory={this.state.filterCategory}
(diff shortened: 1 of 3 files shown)


lauramosher (changes requested):
I suspect that we are passing more than we need to for `user` to the front end. This works fine (though one small change is needed before we can merge); however, we may want to instead explicitly return the attributes we need:

```
json.user do
  json.(current_user, :role, :display_name, :other_attrs)
end
```

rudokemper:
> I suspect that we are passing more than we need to for `user` to the front end. This works fine (though one small change is needed before we can merge); however, we may want to instead explicitly return the attributes we need:
> 
> ```
> json.user do
>   json.(current_user, :role, :display_name, :other_attrs)
> end
> ```

thanks, that is a smarter way of doing it. adapted the code.

lauramosher (approved):
🙌

PR #851, opened 2022-10-12
Docker setup improvements

lauramosher:
Adds a one-stop script to setup terrastories for offline and/or local development.

Updates Rails 6 to latest patch, removes spring, and fixes annotations.


rudokemper:
Thanks @lauramosher, this is so great.

Having played with this, I think there is just one piece of the puzzle missing: setting things up so Terrastories can be served on a WiFi hotspot via a local device (as shown [here](https://acteam.sharepoint.com/:b:/g/ACTUS/mapping/EbzGbP3dSDVPmmTsKToYlLQBA72Tza8cWKWsA7KzTlMB0w)).

I reviewed and went through our [old notes](https://gist.github.com/kalimar/ed14b5d026220ee5cd81d416b4f67b7b) and got this working with the current setup by setting up terrastories.local as 192.168.0.1 (instead of 127.0.0.1) in `etc/hosts`, and also adding `address1=192.168.0.1/24` in the ipv4 section of `/etc/NetworkManager/system-connections/Hotspot`. 

It seems we could adapt the setup script to ask "Do you intend to serve Terrastories via a Hotspot connection? This means that any device connected to your Field Kit device can load Terrastories in the browser once connected to the Hotspot." and then set the ip to either 192.168.0.1 or 127.0.0.1 depending on the answer. (Plus adding that address1 line to the `Hotspot` file if Y).

Or is there a way to have both? So `terrastories.local` will always work both locally AND if served on a Hotspot.

Hopefully that's a quick and painless addition, but you might think about something I'm not considering Laura. Also happy to keep this as a future issue to be resolved since we have other stuff to work on :)

rudokemper:
Related to that, is there any harm in setting all of the Docker services to `restart: always` for the offline profile? The db service does not auto restart when you turn off/on a Terrastories field kit.

lauramosher:
Hey @rudokemper! I reviewed the setup notes and the NUC instructions.

I think we need to start thinking about "Offline Mode" for _development_ and Offline Mode for _field kit installation_. The script here is intended to get Terrastories built in standard or offline mode for feature development / ensure that it all works before it gets deployed to our online server or an actual NUC. While it does do several of the steps needed to install on a NUC, configuring network pieces is not exactly an application specific requirement for development.

Plus, once we start creeping into Network territory, things get less..scriptable. Network configuration varies based on the OS — setting up a hotspot / adding an entry to ipv4 (or ipv6) on a Mac is vastly different than on a Linux machine. And even with linux, it can vary by distro.

Thinking about this more, it may be worthwhile for us to not specify a local domain in offline development mode (e.g. terrastories.local) or modify the /etc/hosts file. If you skip that step, localhost at port 80 works just fine with nginx-proxy; and when specified, a local network hostname (machine name or the using ngrok etc) would also work without modification to the /etc/hosts file.

For those that *do* want to have a local domain, adding it to /etc/hosts or configuring a local hotspot could be a separate how-to/walk through tutorial.

FWIW: I'm good with leaving this as-is since most will be developing in online mode.

> Related to that, is there any harm in setting all of the Docker services to restart: always for the offline profile?

No, I don't think there is any issue with this.

rudokemper:
Thanks @lauramosher, that all makes sense. 

I do agree that defining `terrastories.local` does not really seem necessary for the local development script setup. And I think the idea of having two separate offline workflows (for dev and field kit installation) makes good sense.

Hear you on the difficulties of scripting for different networks. Maybe we script what we can, and add any instructions about e.g. setting up hotspot, stable ip address, to the script for the user to manually configure from there. As well as documenting the whole process in a guide page in the upcoming Gitbook that I plan to work on for Terrastories.

Shall we go ahead and consider any further development on an offline field kit workflow as a separate bucket of work, and ship this?

lauramosher:
Yes, lets! :D

rudokemper (approved):
🥇 amazing script. Huge added value to the local deployment process!

PR #852, opened 2022-10-13
Administrate Refactor Part 2: Switch to new dashboard!

lauramosher:
In part 1 of the refactor, we added a new member and super-admin dashboard. This was added alongside the original Administrate admin CMS interface.

In Part 2, we now use the new dashboard by switching over the links. Administate cms will be still be accessible by accessing /admin instead of /member if necessary.


rudokemper (approved):
🚢  let's do it!

PR #861, opened 2022-10-25
Add PT translations

julinvictus:
For `rails/config/locales/pt/devise.pt.yml` file, I changed the keys' order to keep in sync with `devise.en.yml`


lauramosher (approved):
Thank you!

PR #866, opened 2022-10-27
Administrate refactor part 3: remove it!

lauramosher:
Removes all of the old /admin routes, administrate gems, configuration, and all of the styles, assets, views, dashboards, helpers, etc.

I *believe* this is the administrate code, but we may find some lingering things as we keep cleaning up.

Note: I copied any "helpers" that we had defined in the `administrate.*.yml` locale files and put them into `models.*.yml` just so we kept their reference. I also added fallbacks for locales. `:mat` is specified to fallback to `:nl`, but everything else will fallback to `:en` now.


rudokemper (approved):
Tested this out on staging, local dev, local offline, clicked on all there is to click 😆 and no broken links to Administrate detected anywhere.

Let's rip it out!! 🔪

PR #870, opened 2022-11-01
Chore: Namespace SuperAdmin controllers

lauramosher:
This PR separates the super admin dashboard area of the CMS into its own namespace.

### So now:

Member routes (including Community Admins) are found in the Dashboard namespace.
Super Admin routes (setting up and managing online Communities) are found in the SuperAdmin namespace.

### And route-wise:

Member routes are found at /member
Super Admin routes are found at /admin

### In Production / online mode:
 
If you are a member and attempt to navigate to /admin, you'll reach a 404.
If you are a super admin and attempt to navigate to /member, you'll also reach a 404.

I haven't tested offline field kit mode yet; so I'm not sure how nginx will handle no-route-found. In development mode, it displays a RoutingError rails screen.


rudokemper:
This is a good idea, Laura! We will start to have more routes so makes sense to separate these.

I can confirm that the offline mode also yields a RoutingError rails screen (when trying to access /member/ as admin and vice versa).

lauramosher:
Okay, I think that solidifies that we need to separate offline mode into its own environment. Let's plan on tackling that in a separate PR / set of work since that error will occur even without this change for offline mode hitting routes that are not defined.

rudokemper:
Not sure if you were done @lauramosher so didn't merge yet.

PR #871, opened 2022-11-02
chore: add users to metrics dashboard

rudokemper:
Just a small PR to add Users to the super admin metrics dashboard.

image


lauramosher (changes requested):
You will also need to update the communities/show.html.erb to send `users_count` as well, or wrap the user's count similar to the communities count in the partial to ensure that it doesn't error when you look at a specific community dashboard.

rudokemper:
Oops, forgot the metrics appeared there as well.

lauramosher (approved):

PR #874, opened 2022-11-09
chore: add m4a mime type

rudokemper:
One of our users tried to add an m4a audio recording (from iPhone) to `name_audio` in the Place model, and she wasn't able to because we currently only permit mpeg and wav.

This is a quick PR to add m4a mime type to media attachments for stories and places (place name).


lauramosher (approved):
Works for me!

PR #878, opened 2022-11-11
Feat: Mark Community for Public Exploration

lauramosher:
### ⚠️ This feature is behind a feature flag!!
(and also doesn't really do anything outwardly, I'm just trying to keep PR's small and feature flags allow us to build more incrementally)

This adds the ability for communities to opt-in to their Community being publicly explorable.

At the point, opting in does not serve any purpose, so this feature is behind a feature flag until more of the feature can be built out.

## Testing

1. Log in as a Super Admin and create the `public_communities` feature.
2. Enable it for your preferred community or the `developer` group which will enable the `public_communities` feature for any community that matches "Ruby for Good"
3. Log out, and log back in using the Admin account for the enabled community.
4. Note: There should be a new Admin link to "Settings"
    image
5. Communities can now enable Beta Access and/or Public Exploration
    image
6. Note: There is also a secondary flag, `split_settings` which can be enabled separately or in conjunction — for now enabling this but not public communities will simply provide the new link and the ability to enable Beta access.

### DIFF
--- a/rails/app/controllers/dashboard/communities_controller.rb
+++ b/rails/app/controllers/dashboard/communities_controller.rb
@@ -8,13 +8,16 @@ module Dashboard
       @community = authorize current_user.community
 
       @community.update(community_params)
+
+      redirect_to community_settings_path
     end
 
     private
 
     def community_params
       params.require(:community).permit(
-        :beta
+        :beta,
+        :public
       )
     end
   end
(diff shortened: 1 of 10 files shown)


rudokemper (approved):
Looks good Laura. The only difference I saw from test-driving this PR, is that just enabling `public_communities` does add the Settings menu for an admin user of a communities that opted in, but clicking on it yields a message that the user does not have authority to access. Once I enabled `split_settings`, however, I was able to access the menu and toggle the features.

lauramosher:
lol it would help if I didn't spell the feature wrong 🙈

PR #879, opened 2022-11-11
fix: error creating user without an email

lauramosher:
When we added the ability to sign in with a username, we made email optional. Except that if more than one user was created without an email, it would fail a database uniqueness constraint.

## Changes

- Removes database unique constraint on `users_email` index. This migration is irreversible (once the change is made it stays), however, it's setup so if we need to rollback, the migration will allow us to do that.
- Removes the unused Devise configuration and views such as Confirmable, Lockable, and Recoverable. This also disables Registerable — however the code remains in case we wish to utilize this in the future.
- Fixes an issue where if a non-admin User was updating their profile, they'd receive an alert stating that they couldn't perform that action — this was because it was attempting to redirect them to the Admin user view which they cannot do, and not because the change failed to save. Now, if the user is a non-admin, it'll redirect them back to their profile page. For an Admin, it will check to see where they were referred from (profile) before redirecting them back to either their profile page or to the admin user route.
- Fixes an issue where a Viewer User could view the available Speakers from anonymous stories in the list, but not click on one to view them in depth.


rudokemper (approved):
👍👍👍

PR #896, opened 2022-12-19
chore: do not show user profile path for super admin

rudokemper:
This wraps the `user_profile_path` in a conditional that checks if the user is not super admin.

### DIFF
--- a/rails/app/views/layouts/dashboard.html.erb
+++ b/rails/app/views/layouts/dashboard.html.erb
@@ -86,7 +86,7 @@
             <% end %>
             <ul class="dropdown" aria-label="submenu">
               <% if !current_user.super_admin %>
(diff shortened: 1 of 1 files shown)


lauramosher (approved):

rails/app/views/layouts/dashboard.html.erb
```diff
@@ -85,9 +85,9 @@
               <%= image_tag asset_path("speaker.png"), class: "img img--rounded img--micro" %>
             <% end %>
             <ul class="dropdown" aria-label="submenu">
-              <% if current_user.super_admin %>
+              <% if !current_user.super_admin %>
+                <li><%= link_to t("profile"), user_profile_path %></a></li>
```
lauramosher, line 89:
```suggestion
                <li><%= link_to t("profile"), user_profile_path %></li>
```

Issue #856, opened 2022-10-15
[Bug] When a new language is added to Terrastories, values don't show on the front end until another change is made to i18n


rudokemper:
Terrastories is using Ruby i18n to provide languages. These are added to `rails/config/locales`.

There is a short bit of code in `rails/config/application.rb` that automatically sets the available locales to whatever is found in that directory. 

However, when you first add a new language to Terrastories, it seems to not load the values (from `en.yml` or equivalent) on the front end at first, as per the screenshot below. It _does_ load everywhere else in the application (e.g. devise, administrate, dashboard, model, as well as the welcome page which also loads values from `en.yml` or equivalent).

Then, once you make a change to _any_ of the locales, it _will_ load the front end translations for that new language.

Let's solve this problem so that the front end loads the new locale values right away.

Steps to reproduce:
1. Make sure that you've started Terrastories at least once.
1. Copy the `en` folder in `locales` and name it any other language acronym; this can be arbitrary, but you could use, say, Italian with `it`.
2. Replace `en` to `it` in all of the filenames, as well as in the first line of each file.
3. Start Terrastories. You should see the "missing translation" messages on the front end (`/home/`) despite being set properly in the `it.yml` file.

![Missing translations bug](https://user-images.githubusercontent.com/31662219/195993003-b7e54c35-5719-4a3a-8786-e19e8f760727.jpg)

rudokemper:
You can see this with the recently merged Aramaic translations in the master branch.

![image](https://user-images.githubusercontent.com/31662219/196582200-e0763c53-142b-4cf7-9465-30000ba0d0e8.png)

DuldR:
If this is still available, i'd like to take a crack it!

rudokemper:
Thanks @DuldR ! I assigned you.

DuldR:
@rudokemper 

> Then, once you make a change to _any_ of the locales, it _will_ load the front end translations for that new language.

I can't seem to get this portion to reproduce. What changes can I make to the other locales so that it will load correctly? Do you mean in the `yml` files?

rudokemper:
@DuldR yes in the `yml` files. Update any of the translations in the existing yml files. It should "fix" the problem of the missing ones.

PR #901, opened 2023-01-25
Include new translations by clearing cache and precompiling assets

DuldR:
Fixes [#856](https://github.com/Terrastories/terrastories/issues/856) 

Currently, we're relying on the `i18n-js` gem and utilizing it through the asset pipeline for our translations. When adding a new locale, these changes aren't being reflected on the front end. Only after changing any of the existing `yml` files will the fingerprint for the assets be updated and trigger a refresh. 

After doing some digging, this is actually the intended design by Sprockets.

Relevant issue: [#213 within i18n-js](https://github.com/fnando/i18n-js/issues/213)
Provided Workarounds: https://github.com/fnando/i18n-js/tree/v3#missing-translations-in-precompiled-files-after-adding-any-new-locale-file

For this PR, I went with the first option of calling rake tasks on the rails container after it's spun up. These commands clear the cache and force a precompile of all the assets. Because the cache has been cleared, it picks up all existing locales(old and new) and provides them to the front end. I've tested on both `offline` and `dev` profiles with the following steps:

1. `bin/setup`
2. `docker compose --profile dev up` or `docker compose --profile offline up`
3.  Load `localhost:3000` and confirm splash page with translations
4.  Add a new locale: inside of `config/locales`, `cp -r en/ it/` then modify all `yml` files within the new directory from the `en` to `it` key
5. Kill  `docker compose` with ctrl+c and rerun the command from step 2
6. Load `localhost:3000` and new it translations will be present

While this won't "hot load" new translations, it will ensure anything new will be included at the start of the rails container. 

Option 2 from the `i18n-js` docs  is what we're already doing.
Option 3 from the `i18n-js` docs might be an easier and better solution, but wanted to discuss on path forward.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -26,6 +26,7 @@ services:
       - 3000:3000
     expose:
       - "3000"
+    command: sh -c "bundle exec rake tmp:cache:clear && bundle exec rake assets:precompile && bundle exec rails s -b '0.0.0.0'"
     environment:
       - CHECK_BUNDLER=yes
     volumes:
@@ -64,6 +65,7 @@ services:
       - 3030:3030
     expose:
       - "3030"
+    command: sh -c "bundle exec rake tmp:cache:clear && bundle exec rake assets:precompile && bundle exec rails s -b '0.0.0.0'"
     volumes:
       - bundler:/usr/local/bundle
       - ./rails:/api


rudokemper:
Thanks for this work @DuldR, and sorry for the long delay in getting back to this. But we ended up handling our i18n differently and are no longer using `i18n-js`, so we have to close this one. Nevertheless, thank you for putting in this work and research.

PR #903, opened 2023-01-31
Fix: import

lauramosher:
Somewhere between coding the import and upgrading some dependencies, this broke spectacularly. 😭 

So now, there are tests to hopefully catch these edge cases. This definitely fixes the issue that one of our team encountered during a demo. This also uncovered and fixed a few other issues relating to importing of attachments and creating associated records.


rails/spec/models/place_spec.rb
```diff
@@ -6,6 +6,118 @@
...
+    }}
+
+    it "raises HeaderMismatchError when mapped headers are missing" do
+      expect {
+        described_class.import(
+          file_fixture('place_without_media.csv'),
+          community.id,
+          mapped_headers.merge({photo: "other"})
+        )
+      }.to raise_error(Importable::FileImporter::HeaderMismatchError)
```
albertchae, line 46: maybe a future improvement here is to add which headers were mismatched to the exception message?

rails/spec/models/place_spec.rb
```diff
@@ -6,148 +6,115 @@
...
-    let(:community) { FactoryBot.create(:community) }
-
-    it 'should return a geoJson point' do
-      fixture_data = file_fixture('place_with_media.csv').read
-      described_class.import_csv(fixture_data, community)
-      place = described_class.first
-      expect(place.point_geojson.keys).to include('properties')
+    context "when row is invalid" do
+      it "returns invalid rows with errors" do
+        import = described_class.import(file_fixture('invalid places.csv'), community.id, mapped_headers)
```
albertchae, line 113: looking at this fixture file, I'm not sure why it's invalid. Could you tell me why? 

EDIT: Oh, is it because the long/lat are out of range?

rails/spec/models/story_spec.rb
```diff
@@ -26,62 +28,203 @@
...
+          date_interviewed
+          language
+          interview_location_id
+          interviewer_id
+          topic
+          speakers
+          places
+          media
+        ]
+      )
```
albertchae, line 47: A question after seeing a few of these tests – does the order matter here? I'm guessing no because the `mapped_headers` hash is used to parse the CSVs right?

a small nit if it turns out order doesn't matter, I usually like to use https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/contain-exactly-matcher instead of exact array equals. I don't think this change is needed for this PR but maybe if we run into issues maintaining these tests we can introduce it.

rails/spec/models/theme_spec.rb
```diff
@@ -1,32 +1,28 @@
...
-    expect(theme).to be_valid
-  end
+  let(:theme) {FactoryBot.create(:theme, community: FactoryBot.create(:community))}
 
   it "raises an error with invalid center latitude" do
     expect do
       theme.center_lat = 100
       theme.save!
-    end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Center lat value should be between -90 and 90")
+    end.to raise_error(ActiveRecord::RecordInvalid, "Validation failed: Latitude value should be between -90 and 90")
```
albertchae, line 10: ok this answers my other question 😆

albertchae (approved):
LGTM, all tests pass and make sense.

A meta comment I had is it wasn't always very obvious why a particular CSV was suitable for a test case but I think that's mostly due to my unfamiliarity. It's also hard to avoid when testing CSVs anyway since there is so much data...

rails/spec/models/place_spec.rb
lauramosher: > EDIT: Oh, is it because the long/lat are out of range?

Yup!

rails/spec/models/story_spec.rb
lauramosher: Ah yeah, good call! I don't believe we care about what order the headers are returned. They are extracted from the database/rails so there is some intrinsic "ordering" but that also may be different on different machines.

PR #904, opened 2023-01-31
[Draft] Allow super admin to reset passwords for community admin users

### LINKED ISSUE #802: [Rails] Allow super admin to reset passwords for community admin users
When logged in as a super admin (e.g. via username `super@terrastories.io` and password `terrastories` per the default seed data), we are able to view the communities that are located on the Terrastories instance, as well as the users registered to those communities. Let's make it possible for the super admin to reset a password for any of the users on a community with `admin` credentials.

User story: if a Community administrator loses their password, the super admin should be able to reset that for them.

DuldR:
Fixes #802

Related PR: https://github.com/Terrastories/terrastories/pull/905

I wasn't sure on how powerful we want a super admin to be, so I've essentially allowed them the max permissions for this PR. (More discussion: https://rubyforgood.slack.com/archives/CAH9XQX2A/p1675975728042129?thread_ts=1675971543.153629&cid=CAH9XQX2A)

Updated this PR to only allow a `super_admin` to update a communities admin password. 

Community User Page
image

Super Admin Reset Password Page
image

Admin Reset Password Page
image


rails/app/controllers/super_admin/users_controller.rb
```diff
@@ -0,0 +1,33 @@
+module SuperAdmin
+  class UsersController < ApplicationController
+    def edit
+      @user = User.find(params[:id])
+
+      if @user.role == 'admin'
```
lauramosher, line 6: Since the roles are handled by an enum, we can use the built-in helper methods to check the role:

```suggestion
      if @user.admin?
```

rails/app/controllers/super_admin/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+        @user
+      else
+        flash[:alert] = 'You are not authorized to perform this action.'
+        redirect_to(root_path)
```
lauramosher, line 10: I'm wondering if it would be better for us to redirect to the community path rather than to the root path which shows the metrics for all of the communities for the super admin. Since it seems the current path is: Metrics root page (after login) -> Communities -> Community -> Find the user in the tiny list -> edit. On save, going all the way back feels jarring.

Alternatively, it may also be worth creating a new list view under the Super Admin dashboard that lists all of the admin users in the system that we can use to redirect to and make it easier for super admins to reset admin passwords, regardless of the community.

@rudokemper thoughts?

rails/app/controllers/super_admin/users_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      end
+    end
+
+    def update
+      @user = User.find(params[:id])
+      if @user.role != 'admin'
+        redirect_to(root_path)
```
lauramosher, line 17: Let's add a flash alert here to denote why it failed/they are being redirected.

rails/app/views/super_admin/users/edit.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+    <%= @user.display_name %>
+    <div class="small">
+      <span class="badge"><%= User.human_attribute_name("role.#{@user.role}") %></span>
+    </div>
+  </h2>
+  </div>
+<% end %>
+
+<%= form_with url: member_path(@user), model: @user, multipart: true, class: "form aligned", data: {remote: false}, id: "userForm", locale: true do |f| %>
```
lauramosher, line 36: We can remove `multipart` from this form since there are no file uploads.

rails/spec/requests/admin_user_request_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'rails_helper'
+
+RSpec.describe "SuperAdmin user request", type: :request do
+  describe "GET edit" do
+    context "with a super admin privileges" do
+      let(:super_admin) { FactoryBot.create(:user, role: 100, super_admin: true) }
+      let(:user) { FactoryBot.create(:user, email: "user@mail.co", username: "user_name", role: 2) }
```
lauramosher, line 7: In case you were unaware, you can reference the roles by their symbols. So, rather than having to integer map `role: 2`, you can use `role: :admin`.

rails/spec/requests/admin_user_request_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      before do
+        login_as super_admin 
+      end
+
+      it "should return success if the user is an admin" do
```
lauramosher, line 13: Let's also add a spec for when the logged-in user is a super admin, but the user itself is not an admin. The controller accounts for this with a redirect, but that is not encapsulated in a spec here.

rails/spec/requests/admin_user_request_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      before do
+        login_as super_admin 
+      end
+      it "should return success if the user is an admin" do
+        put "/admin/users/#{user.id}", :params => {:user => { password: "securePassword" } }
+
+        expect(response).to have_http_status(302)
+        expect(flash[:notice]).to match('User successfully updated')
+      end
```
lauramosher, line 49: Let's also add a failure spec here for when the user being updated is not an admin here.

rails/app/controllers/super_admin/users_controller.rb
rudokemper: Sorry I missed this back in May. I think the first solution makes the most sense.

rudokemper:
Just saw this again! Wondering what we can do to finish this PR? @DuldR are you still able to work on it, or do you want to leave it here for someone else to finish the remaining tasks as indicated by @lauramosher?

DuldR:
@rudokemper Yes sir, sorry, I can't finish it up. If anyone else wanted to take a whack at it, that'd be great.

PR #905, opened 2023-02-02
Allow super admin to reset passwords for community admin users - Alternate method

### LINKED ISSUE #802: [Rails] Allow super admin to reset passwords for community admin users

DuldR:
Fixes #802

Related PR: https://github.com/Terrastories/terrastories/pull/904

For this functionality, I've changed the authorization logic to allow for a super admin to touch the user edit, show and update methods. Instead of creating it's own super_admin route, this hooks into the existing route available.

I think this methodology is a bit cleaner as the `super_admin` hits the same edit page as an admin and user. So if we ever update the logic anywhere in the pipeline, we'll only need to make a single change instead of keeping two flows in sync.

Super Admin Panel
image

I ensured that a user cannot update any other user with this implementation. An unauthorized error is raised on any attempt to go to another user's page.
image


rails/app/controllers/dashboard/users_controller.rb
```diff
@@ -34,11 +34,11 @@ def create
     end
 
     def show
-      @user = authorize community.users.find(params[:id])
+      @user = authorize User.find(params[:id])
```
albertchae, line 37: one thing that doing the find off a `Community`'s `users` association is that it was implicitly a permissions check too, because `community` is getting pulled from the `current_user` here https://github.com/Terrastories/terrastories/blob/62443134bac8662c8bc2df76e36596b7f81453f0/rails/app/controllers/dashboard/application_controller.rb#L27-L30

If you remember from our discussion, the reason we have to remove it here is that a super admin doesn't belong to a single community, so if we remove this implicit community check, we'd probably have to add an explicit one in `UserPolicy`

rails/app/policies/user_policy.rb
```diff
@@ -11,7 +11,7 @@ def index?
   end
 
   def edit?
-    current_user.admin? || (
+    current_user.super_admin || current_user.admin? || (
```
albertchae, line 14: should this be `super_admin?` (I think the predicate method comes from AR enums)
```
pry(main)> User.last.method(:admin?).source_location
  User Load (11.5ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> ["/usr/local/bundle/ruby/2.5.0/gems/activerecord-6.1.7/lib/active_record/enum.rb", 235]
```

What's weird is that there is a `super_admin` boolean on user too, but at least in my seed file I end up with this:

```
[12] pry(main)> User.last
  User Load (2.6ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> #<User id: 6, email: "public@terrastories.io", created_at: "2023-02-03 03:05:08.225443000 +0000", updated_at: "2023-02-03 03:05:08.225443000 +0000", role: "super_admin", community_id: nil, super_admin: false, username: "terrastories-public", name: nil>
[13] pry(main)> User.last.super_admin
  User Load (0.9ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> false
[14] pry(main)> User.last.super_admin?
  User Load (0.6ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" DESC LIMIT $1  [["LIMIT", 1]]
=> true
```

rails/config/routes.rb
```diff
@@ -1,7 +1,7 @@
 # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
 Rails.application.routes.draw do
   scope '(:locale)', locale: /#{I18n.available_locales.join("|")}/ do
-    scope '/member', module: 'dashboard', constraints: RoleRoutingConstraint.new { |user| !user.super_admin } do
+    scope '/member', module: 'dashboard', constraints: RoleRoutingConstraint.new do
```
albertchae, line 4: Just want to call out that this change makes the assumption that a `super_admin` should be able to access every page, which seems reasonable to me but maybe was intentionally restricted in the past for a reason we're not aware of.

rails/app/controllers/dashboard/users_controller.rb
albertchae: Oh sorry, I wasn't clear what the consequence of this is. I believe without this implicit `community` check, an `admin` of any `community` can access these routes for **other** communities.

rails/app/policies/user_policy.rb
```diff
@@ -11,17 +11,15 @@ def index?
   end
 
   def edit?
-    current_user.admin? || (
-      current_user == user
-    )
+    current_user.super_admin? || (current_user.admin? && (current_user.community_id == user.community_id))
```
albertchae, line 14: I believe you removed the possibility for a user to edit themselves, which is what `current_user == user` enabled

albertchae: I think it'd also be good to name the concept you're using here and factor out a separate method
```ruby
def community_user_admin?
   current_user.admin? && (current_user.community_id == user.community_id)
end
```

rails/app/policies/user_policy.rb
```diff
@@ -11,17 +11,15 @@ def index?
...
   end
 
   def update?
     edit?
   end
 
   def show?
-    current_user.admin?
+    edit?
```
albertchae, line 22: Hm, that's surprising that a user couldn't call show on themselves though 🤔

rails/app/policies/user_policy.rb
DuldR: Yep, you're right. Updated.

rails/app/policies/user_policy.rb
DuldR: You're right! Updated.

rails/app/controllers/dashboard/users_controller.rb
DuldR: Yep, you're right! Added logic and tests to confirm the admin is within the same community.

lauramosher:
Closing this PR in favor of the implementation in #904

PR #910, opened 2023-04-29
Update Gemfile ruby version to 2.7.8

albertchae:
It looks like the ruby:2.7-alpine image was updated to 2.7.8 in https://github.com/docker-library/ruby/commit/7117899e1b7d3d8230fe8021acf76ead8d2f5230
since terrastories was updated

Fixes this error
```
 => ERROR [devcore 5/6] RUN bundle config set --local path /usr/local/bundle &&     bundle config build.nokog  2.1s
------
 > [devcore 5/6] RUN bundle config set --local path /usr/local/bundle &&     bundle config build.nokogiri --use-system-libraries &&     bundle install:

0 2.032 Your Ruby version is 2.7.8, but your Gemfile specified 2.7.7
```

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.7.7'
+ruby '2.7.8'
 
 # lock mail version (required by rails)
 # until upgrade to Ruby 3 / Rails 7


albertchae:
Not sure why running `bin/setup` after this change didn't update the ruby version in `Gemfile.lock` too...

lauramosher:
I think we hit a weird ordering problem.

Since our `Gemfile.lock` file is locked to 2.7.8, you aren't able to bundle with the alpine image. If you update the `Dockerfile` to `FROM ruby:2.7-alpine` to `FROM ruby:2.7.7-alpine` and then run `bin/setup`, does it work for you? (note that there are multiple FROM directives.

And then once we are good there, you can then update the version to 2.7.7 in both the Dockerfile and Gemfile to the latest 2.7.8 version.

I suppose we will need to ensure we always include the patch so our lock file stays in sync.

lauramosher (approved):
🎉

PR #911, opened 2023-04-30
Foreman/Procfile fixes for offline mode

albertchae:
- Respect PORT from compose.yaml in Procfile.dev
  with `unset PORT`, we were ignoring the localweb 3030 port for offline mode, so we'd be stuck on the nginx loading screen because rails was running on port 3000
- Add foreman in gemfile and use bundle exec to run it
  Without this change, we run into this error when trying to run offline mode without internet for the first time

```
terrastories-localweb-1     | Installing foreman...
terrastories-localweb-1     | ERROR:  Could not find a valid gem 'foreman' (>= 0), here is why:
terrastories-localweb-1     |           Unable to download data from https://rubygems.org/ - SocketError: Failed to open TCP connection to rubygems.org:443 (getaddrinfo: Name does not resolve) (https://rubygems.org/specs.4.8.gz)
terrastories-localweb-1     | ./bin/dev: exec: line 8: foreman: not found
```

### DIFF
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -87,6 +87,8 @@ group :development do
 
   # annotate models
   gem "annotate"
+
+  gem "foreman"
 end
 
 group :test do
--- a/rails/Gemfile
+++ b/rails/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 
-ruby '2.7.7'
+ruby '2.7.8'
 
 # lock mail version (required by rails)
 # until upgrade to Ruby 3 / Rails 7
--- a/rails/Gemfile.lock
+++ b/rails/Gemfile.lock
@@ -139,6 +139,7 @@ GEM
     flipper-active_record (0.26.0)
       activerecord (>= 4.2, < 8)
       flipper (~> 0.26.0)
+    foreman (0.87.2)
     formatador (1.1.0)
     geocoder (1.8.1)
     globalid (1.1.0)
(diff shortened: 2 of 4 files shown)


rails/Gemfile.lock
```diff
@@ -331,7 +330,7 @@ GEM
     zeitwerk (2.6.7)
 
 PLATFORMS
-  ruby
+  aarch64-linux-musl
```
albertchae, line 333: not sure where this change is coming from

rails/Gemfile.lock
```diff
@@ -374,7 +374,7 @@ DEPENDENCIES
   webpacker (~> 5.0)
 
 RUBY VERSION
-   ruby 2.7.7p221
+   ruby 2.7.8p225
```
albertchae, line 377: Looks like this updated here finally https://github.com/Terrastories/terrastories/pull/910#issuecomment-1528829225

rails/Gemfile
```diff
@@ -87,6 +87,8 @@ group :development do
 
   # annotate models
   gem "annotate"
+
+  gem "foreman"
```
lauramosher, line 91: Per `foreman`'s installation instructions, this gem [should NOT be installed in the Gemfile](https://github.com/ddollar/foreman#installation).

rails/bin/dev
```diff
@@ -1,8 +1,3 @@
 #!/usr/bin/env sh
 
-if ! gem list foreman -i --silent; then
-  echo "Installing foreman..."
-  gem install foreman
-fi
-
-exec foreman start -f Procfile.dev "$@"
```
lauramosher, line 8: This setup was done in this manner for a reason. See the `foreman` gem notes above.

lauramosher (changes requested):
Let's remove the `foreman` gem, delete the lock file, and re-bundle.

Removing the unset on the PORT seems like a good idea. Ideally, foreman and the dev Procfile are only used by developers working on offline mode or standard. Official offline setup should be done using the appropriate flags so the Procfile is not being used. Either way, though, the removal seems like a good idea though!

PR #912, opened 2023-05-01
Fix Offline Mode Configuration & Setup

lauramosher:
These are a few of the issues that some of our community admins have been running into while setting up with Docker & Docker Compose.

1. The default DATABASE_URL was using docker's internal host rather than the service name as host.
2. `SECRET_KEY_BASE` was missing. RAILS_ENV=offline now correctly generates this on first-boot.
3. The official Ruby 2.7 docker image was updated with a minor patch, breaking our builds. We now lock the minor patch.
4. Prefix and precompile assets when booted into RAILS_ENV=offline mode.

These changes "break" (loosely) using the offline environment for development testing. Instead, it's recommended for developers enhancing offline mode to utilize development environment and set the OFFLINE_MAP_STYLE to the local running tileserver using the fully-qualified URL to the desired style (such as localhost:8080/styles/terrastories-map.json).

This isn't perfect, but it should unblock our members who are unable to set up with the latest version on GitHub.

I was able to successfully test setting up with the `bin/setup` script and doing no additional configuration adjustments.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -136,8 +136,18 @@ docker compose up -d db
 ## build app
 docker compose --profile $profile build
 
-## setup and seed database
-docker compose --profile $profile run --rm $container bin/rails db:migrate db:seed
+## setup database
+docker compose --profile $profile run --rm $container bin/rails db:migrate
+
+## Seed database with demo data
+if [[ "$profile" = "dev" ]]; then
+  docker compose --profile dev run --rm $container bin/rails db:seed
+fi
+
+## Precompile assets for offline mode
+if [[ "$profile" = "offline" ]]; then
+  docker compose --profile offline run --rm -e RAILS_ENV=offline $container bin/rails assets:precompile
+fi
 
 ## run yarn
 docker compose --profile $profile run --rm $container yarn
--- a/rails/Dockerfile
+++ b/rails/Dockerfile
@@ -8,7 +8,7 @@
 # == Development Image
 #
 # Also used as the base for building assets for prod images.
-FROM ruby:2.7-alpine as devcore
+FROM ruby:2.7.7-alpine as devcore
 
 RUN apk --no-cache add --update \
     build-base \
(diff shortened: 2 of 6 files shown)


rails/Dockerfile
```diff
@@ -8,7 +8,7 @@
 # == Development Image
 #
 # Also used as the base for building assets for prod images.
-FROM ruby:2.7-alpine as devcore
```
albertchae, line 11: Should I close https://github.com/Terrastories/terrastories/pull/910? Or should I update that PR to be for updating from 2.7.7 to 2.7.8?

bin/setup
```diff
@@ -136,8 +136,18 @@ docker compose up -d db
...
 
-## setup and seed database
-docker compose --profile $profile run --rm $container bin/rails db:migrate db:seed
+## setup database
+docker compose --profile $profile run --rm $container bin/rails db:migrate
+
+## Seed database with demo data
+if [[ "$profile" = "dev" ]]; then
+  docker compose --profile dev run --rm $container bin/rails db:seed
+fi
```
albertchae, line 145: Do we not want to seed the database in offline because we're assuming it's being used for a field deployment? I saw that the user gets dropped into a setup wizard on loading the site for the first time.

albertchae (approved):
Tested on my machine and it worked for me. Thanks!

BTW I noticed that the port for localweb was starting up on 3030 properly. Did one of your changes here somehow make it not run `bin/dev`?

rails/Dockerfile
lauramosher: Let's keep it and have it update from 2.7.7 to 2.7.8!

bin/setup
lauramosher: Yes, that's correct. I think I thought to leave it seeding until we were ready to actually use the onboarding flow, but enough folks have tested it now that I think it's worth being the default.

lauramosher:
> BTW I noticed that the port for localweb was starting up on 3030 properly. Did one of your changes here somehow make it not run bin/dev?

Yes, I apparently had the default value for RAILS_ENV= to be `development` instead of `offline` which already correctly switches between using `/bin/dev` versus booting up `/scripts/server`.

PR #916, opened 2023-05-05
Remove offline scope on policy_scope

lauramosher:
Let it handle scoping by permissions; offline mode shouldn't change that.

This was causing runs booted with RAILS_ENV=offline to never allow admins or editors to view speakers or places that didn't have an anonymous story attached.


rudokemper (approved):
Can confirm this resolves the issues reported; thanks!

PR #917, opened 2023-05-18
fix(import): Imports with file attachments failed

lauramosher:
This fixes an issue where imports that contained file attachments were not importing correctly despite green tests. This is because CSV used string headers and my tests were using symbol headers; this caused the tests to pass, but the actual import failed if the import file contained file attachments. Consequently, the import did work if no attachments were present.

This likely worked before — and why our click testing was successful when we QA'd this feature when it was first completed — until we upgraded to a version that no longer implicitly provided `with_indifferent_access` to hashy objects that allowed key comparison between string and symbols to work without additional code. I've updated the tests to use string keys and also added the explicit `with_indifferent_access` to ensure that we are not caught off-guard again.

### DIFF
--- a/rails/app/models/concerns/importable.rb
+++ b/rails/app/models/concerns/importable.rb
@@ -106,7 +106,9 @@ module Importable
 
     def save_importable_records
       importable_rows.each do |row|
-        attributes = row.dup
+        # with_indifferent_access to allow key access as strings or symbols
+        # since CSV uses strings and AR uses symbols.
+        attributes = row.dup.with_indifferent_access
 
         # Remove media from attributes to attach later.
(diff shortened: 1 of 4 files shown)


albertchae (approved):
The changes make sense, so approving the PR.

On a side note I was trying to run the specs locally to verify but forgot how I did that a few months ago and the steps in https://github.com/Terrastories/terrastories/blob/master/documentation/DEVELOPMENT.md (specifically `docker-compose exec web rspec` don't seem to work). How are you running them? I was able to start `sh` (but not `bash`) in the web container and try running rspec from there, but then I get a bunch of errors about FactoryBot loading so I think I'm doing something wrong...

Issue #805, opened 2022-09-29
[CMS] Let's add a location picker for Places using a map


rudokemper:
The newly merged Admin CMS refactor adds a whole new CMS to Terrastories, accessed at `/en/member/` when logged in to a community. In the Themes page, it is now possible to set map settings such as bounding box, zoom level, pitch, and bearing by interacting with an inset map, provided by either Mapbox.com or local tiles served by Tileserver-GL:

![image](https://user-images.githubusercontent.com/31662219/193128219-1e4d581a-2417-4208-b9f4-76babd150551.png)

Currently, the only option to create a Place is to manually enter latitude and longitude coordinates.

![image](https://user-images.githubusercontent.com/31662219/193128427-150d7f8b-44da-4eb4-9e6f-efab5510b0c8.png)

Let's add a similar map to the Place edit view, where the user can pick coordinates by moving around on the map. The behavior for this can be similar to this [example from Mapbox](https://blog.mapbox.com/how-to-build-a-location-picker-for-your-app-8e61be7fc9cc), with a drop pin located in the center of the map, the lat/long of which synchronizes with the lat/long fields in the form.

PR #920, opened 2023-07-06
Fix mapbox-gl.css import

albertchae:
With the `.css` extension, we weren't actually inlining the CSS during
the build process, so we'd see the following issues in offline mode when
visiting a page that uses `dashboard.css` (e.g. http://terrastories.local/en/member/theme/edit)

a 404 for the `mapbox-gl.css` file
```
GET http://terrastories.local/offline-assets/mapbox-gl/dist/mapbox-gl.css net::ERR_ABORTED 404 (Not Found)
```

At the top of `rails/public/offline-assets/dashboard-<DIGEST>.css`, we'd
see this line

```
@import 'mapbox-gl/dist/mapbox-gl.css';
````

There were also some visual glitches which you can see in these screenshots.

Screenshot 2023-07-06 at 2 13 14 PM

I noticed this when trying to set the marker for https://github.com/Terrastories/terrastories/issues/805.
After this change, the 404 is gone, the CSS is properly inlined in
`dashboard-<DIGEST>.css` and no visual glitches.

Screenshot 2023-07-06 at 2 18 41 PM

### DIFF
--- a/rails/app/assets/stylesheets/dashboard.sass.scss
+++ b/rails/app/assets/stylesheets/dashboard.sass.scss
@@ -1,4 +1,4 @@
-@import 'mapbox-gl/dist/mapbox-gl.css';
+@import 'mapbox-gl/dist/mapbox-gl';
 
 @import 'core/reset';
 @import 'core/colors';


rudokemper (approved):
Reproduced the bug locally, and can confirm this fixes the issue. Thanks!

PR #921, opened 2023-07-06
Move the map center after editing the lat/long fields

albertchae:
When dragging the map around, the center coordinate input fields were synced to the
input fields. This PR implements syncing in the reverse direction in case the user wants to edit the coordinates
manually but wants to see where that is on the map visually.

terrastories

I originally tried to listen for [input event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event), but it would fire as the user is typing and modify the value to include the 5 fixed decimal points, which would effectively prevent the user from entering any values more than a single digit. So I am listening for [change event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) instead, which has the small negative that it won't fire until the input field loses focus.

Also there is a small refactor to reuse the variables for the input fields

I noticed this while working on https://github.com/Terrastories/terrastories/issues/805

### DIFF
--- a/rails/app/views/dashboard/themes/edit.html.erb
+++ b/rails/app/views/dashboard/themes/edit.html.erb
@@ -162,6 +162,8 @@
 
 <script>
 const themeForm = document.querySelector("#themeForm");
+const themeCenterLatInput = themeForm.querySelector("#theme_center_lat");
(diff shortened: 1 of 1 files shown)


rudokemper (approved):
Great contribution @albertchae - I didn't even realize that don't already have this behavior in place :)

PR #922, opened 2023-07-06
Implement location picker map for places

albertchae:
Closes https://github.com/Terrastories/terrastories/issues/805

This is a very similar implementation as the one for [themes](https://github.com/Terrastories/terrastories/blob/bb0784bd162dc2095b75afb7d01f0841ee0dadb4/rails/app/views/dashboard/themes/edit.html.erb#L163) (using the same 2 column CSS and set of event listeners), and includes functionality like https://github.com/Terrastories/terrastories/pull/921 as well.

terrastories

I verified this works offline as well, except we need to merge https://github.com/Terrastories/terrastories/pull/920 to fix the CSS issue.

I also verified it works for both Places #new and #edit (they share the _form partial)

As the issue mentioned, I locked the marker to the center of the map
instead of making it independently draggable. One of the consequences of
this is that it's a bit weird to set a point to one of the latitude
extremes (i.e. -90 or 90). Even editing this manually will get the value
reset to the center of the current map viewport depending on the zoom
level.

boundaries

### DIFF
--- a/rails/app/views/dashboard/places/_form.html.erb
+++ b/rails/app/views/dashboard/places/_form.html.erb
@@ -60,7 +60,7 @@ const placeMap = new mapboxgl.Map({
   container: "placeMap", // container ID
   center: coordinates, // starting position [lng, lat]
   hash: false,
-  zoom: <%= @initial_zoom %>,
+  zoom: <%= current_community.theme.zoom || 5 %>, // starting zoom
   style: "<%= current_community.theme.mapbox_style %>", // style URL or style object
   cooperativeGestures: true
 });
--- a/rails/app/views/dashboard/places/_form.html.erb
+++ b/rails/app/views/dashboard/places/_form.html.erb
@@ -60,7 +60,7 @@ const placeMap = new mapboxgl.Map({
   container: "placeMap", // container ID
   center: coordinates, // starting position [lng, lat]
   hash: false,
-  zoom: 3, // starting zoom
+  zoom: <%= @initial_zoom %>,
   style: "<%= current_community.theme.mapbox_style %>", // style URL or style object
   cooperativeGestures: true
 });
--- a/rails/app/views/dashboard/places/_form.html.erb
+++ b/rails/app/views/dashboard/places/_form.html.erb
@@ -1,4 +1,4 @@
-<%= form_with model: @place, multipart: true, data: {remote: false}, class: "form aligned" do |f| %>
+<%= form_with model: @place, multipart: true, data: {remote: false}, class: "form aligned", id: "place_form" do |f| %>
   <%= render partial: "shared/form_errors", locals: { resource: @place } %>
 
   <%= f.label :name, class: "required" %>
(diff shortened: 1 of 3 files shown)


rails/app/views/dashboard/places/_form.html.erb
```diff
@@ -34,11 +34,57 @@
...
+const placeForm = document.querySelector("#place_form");
+const placeLatInput = placeForm.querySelector("#place_lat");
+const placeLongInput = placeForm.querySelector("#place_long");
+const coordinates = [placeLongInput.value, placeLatInput.value]
+const placeMap = new mapboxgl.Map({
+  accessToken: "<%= current_community.theme.mapbox_token %>",
+  container: "placeMap", // container ID
+  center: coordinates, // starting position [lng, lat]
+  hash: false,
+  zoom: 3, // starting zoom
```
rudokemper, line 63: Let's set this to `5`. `3` is probably a little zoomed out for most use cases; with `5` we can still make out a good amount of detail in the default offline vector tiles, but is more focused on a region.

rudokemper (changes requested):
Yeeesssss this is huge! Will make so many user's lives easier. Just requested one small code change for the default zoom level. Thanks @albertchae!

rails/app/views/dashboard/places/_form.html.erb
albertchae: ok no problem, but I think if we make this 5 we should maybe make the initial position something that isn't null island 0, 0 because all you can see is ocean and it's not super obvious you can interact with it. Here's a screenshot

image

rudokemper: Ah, good point. This actually makes me think we're better off using the zoom level and center lat/longs from the community map settings. Null island is not a useful reference for anyone, and users will have to navigate away from it to their area of interest every time.

I'd say let's ship this with zoom level 7 and have that be another ticket (unless you think it's doable with minimal work having just dug into the code for both of these?)

rudokemper (approved):

rails/app/views/dashboard/places/_form.html.erb
albertchae: It was pretty easy to use the community theme's zoom and lat/long. I implemented it in https://github.com/Terrastories/terrastories/pull/922/commits/b8d8f1897aedd30a99b1be4f7c9c94f4c06809ae

rudokemper: Awesome! We just need one small fix, which is to implement the change for the edit form - the Zoom value there is showing as null.

image

albertchae: Oh great catch, I was debating whether to use an instance variable or make it more consistent with the other values pulled off the theme there, but forgot for edit instead of new we wouldn't have access to the instance variable. I pushed a fix for it in https://github.com/Terrastories/terrastories/pull/922/commits/55f88167764a49ffee4c6dad7c498284e293e4c7

I also remembered I should probably handle the case where Theme is missing or has null values for lat/long/zoom. It looks like we should always have a theme thanks to this after_create hook https://github.com/Terrastories/terrastories/blob/ca68805a53313313d8c217768a2973982b430c83/rails/app/models/community.rb#L12, but it does look like it's possible for the other values to be null, so I added fallbacks to 0, 0 with zoom 5. Hopefully most users set up their theme before adding places so this won't be much of an issue.

rudokemper:
Awesome work on this @albertchae, and great idea on adding the fallback! LGTM :raised_hands:

PR #923, opened 2023-07-11
Add mapbox-gl back to react stylesheet

rudokemper:
This PR fixes an issue where the mapbox-gl stylesheet was no longer being accessed by React, resulting in unstyled popups (among other things, probably).

Before:

image

After:

image

### DIFF
--- a/rails/app/assets/stylesheets/react.sass.scss
+++ b/rails/app/assets/stylesheets/react.sass.scss
@@ -1,2 +1,4 @@
+@import 'mapbox-gl/dist/mapbox-gl';
+
 @import 'core/core';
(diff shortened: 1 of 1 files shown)


albertchae (approved):
Oh interesting, do you think this broke because of https://github.com/Terrastories/terrastories/pull/920? Or was it unrelated?

rudokemper:
> Oh interesting, do you think this broke because of #920? Or was it unrelated?

That was my first guess but I think it came from some refactoring work from earlier this year, and it just flew under the radar. #920 helped me figure out why it was broken, pretty much right away :)

PR #927, opened 2023-07-11
Update gem dependencies

lauramosher:
Run `bundle update` for dependency updates


rudokemper (approved):
Did some usual workflows offline and all seems to be working as expected.

Issue #928, opened 2023-07-17
[Bug] CMS search no longer working


rudokemper:
**Describe the bug**
Something that we did since February/March appears to have broken the CMS (Member Dashboard) search in any environment (dev, offline, prod). Any search attempt now leads to the "Oops, sorry about that" error page.

**To Reproduce**
1. Log in to a Terrastories community with any role credential.
2. Access the Member Dashboard.
3. Try searching for anything

Can be reproduced locally on either online or offline dev environment.

![image](https://github.com/Terrastories/terrastories/assets/31662219/069955ee-bafb-403e-8f69-3f94885b9909)

PR #931, opened 2023-07-20
Fix error with missing method community

albertchae:
Fixes https://github.com/Terrastories/terrastories/issues/928

Used current_community which is defined in the grandparent ApplicationController
https://github.com/Terrastories/terrastories/blob/34df5d2aafafd36dd6871bff3b3b90966e0b173d/rails/app/controllers/application_controller.rb#L25-L28

image


rudokemper (approved):
Successfully solves the problem!

PR #935, opened 2023-08-08
feat: wrap story description lines

lauramosher:
When new lines are added to the description (e.g. Shift+Enter), ensure those are appropriately rendered when displaying the description in the story view.

This is done using pure css with white-space render specifications. While there is interest in adding the ability for additional rendering tags (such as links), this is an easy-win with no security risks (injection). I purposely created two separate classnames for the styles in React vs CMS since they are bundled differently and this is likely just a stop-gap toward a more robust implementation.

**In Dashboard**
image

**In Map**
image


rudokemper (approved):
This is great, @lauramosher! I think most, if not all of our users would prefer a white-space render specification over a line break tag.

PR #939, opened 2023-08-22
[WIP] fix: Add scope to Place & Story Model to filter out null coords

rudokemper:
This draft PR adds a scope to the Place and Story Models to permit filtering out any places that have null coordinates (primarily for the front end of TS Main and Explore).

Although we are currently requiring lat/long fields when creating a new Place in the CMS, the Importer currently permits the user to upload Places without lat/longs, hence introducing a bug where Places without lat/long are passed on to React. 

In the future we need to rework the Importer as a whole, but that's a whole 'nother epic.


rudokemper:
This got messy with some accidental rewrites of spec files (my bad) so going to redo the PR.

PR #940, opened 2023-08-29
Chore: Make media its own model

lauramosher:
There are many features that are coming on the roadmap relating to Media attachments to a story. Since most of them will be significantly easier if all of the Media are encapsulated in its own model instead of a `has_many_attached` active storage solution, this PR is all of the chore work to create a standalone Medium model.

This PR:

- [x] Creates a new media table in the database.
- [x] A new Medium model in the app
- [x] Data migration to update the reference of `media` from Story to Media and the new Media to Story (this is reversible)
- [x] Updates all of the references and cleanup to add, remove, and display Media.
- [x] Ensures all references of Media correctly delegate through the Medium model.
- [x] Ensures import works as expected to attach Media attachments.

This PR does NOT:

- Add any of those coming soon roadmap features.
- Address any existing issues with displaying media attachments (if they didn't work before, they still don't work).

### DIFF
--- a/rails/app/controllers/dashboard/stories_controller.rb
+++ b/rails/app/controllers/dashboard/stories_controller.rb
@@ -21,9 +21,12 @@ module Dashboard
 
     def create
       authorize Story
-      @story = community_stories.new(story_params)
+      @story = community_stories.new(story_params.except(:media))
 
       if @story.save
+        story_params[:media].each do |media|
+          m = @story.media.create(media: media)
+        end
         redirect_to @story
       else
         render :new
@@ -41,7 +44,11 @@ module Dashboard
     def update
       @story = authorize community_stories.find(params[:id])
 
-      if @story.update(story_params)
+      if @story.update(story_params.except(:media))
+        story_params[:media].each do |media|
+          m = @story.media.create(media: media)
+        end
+
         redirect_to @story
       else
         render :edit
@@ -59,8 +66,8 @@ module Dashboard
     def delete_media
       @story = authorize community_stories.find(params[:story_id])
 
-      media_blob = @story.media.blobs.find_signed(params[:id])
-      media_blob.attachments.each(&:purge)
+      media = @story.media.find(params[:id])
+      media.destroy
 
       head :ok
     end
--- a/rails/app/models/medium.rb
+++ b/rails/app/models/medium.rb
@@ -21,7 +21,8 @@ class Medium < ApplicationRecord
       # audio types
       :mp3, :aac, :flac, :mp4a, :wav,
       'audio/wav', 'audio/m4a', 'audio/x-m4a', 'audio/x-aac', 'audio/x-flac',
-    ]
+    ],
+    size: { less_than_or_equal_to: 200.megabytes }
 
   delegate :content_type, :blob_id, :blob, to: :media
 end
--- a/rails/app/models/medium.rb
+++ b/rails/app/models/medium.rb
@@ -3,7 +3,11 @@ class Medium < ApplicationRecord
 
   belongs_to :story
 
-  has_one_attached :media
+  has_one_attached :media do
+    def blob_id
+      blob.id
+    end
+  end
 
   validates :media,
     attached: true,
@@ -18,6 +22,8 @@ class Medium < ApplicationRecord
       :mp3, :aac, :flac, :mp4a, :wav,
       'audio/wav', 'audio/m4a', 'audio/x-m4a', 'audio/x-aac', 'audio/x-flac',
     ]
+
+  delegate :content_type, :blob_id, :blob, to: :media
 end
 
 # == Schema Information
--- a/rails/app/models/medium.rb
+++ b/rails/app/models/medium.rb
@@ -1,0 +1,35 @@
+class Medium < ApplicationRecord
+  self.table_name = "media"
+
+  belongs_to :story
+
+  has_one_attached :media
+
+  validates :media,
+    attached: true,
+    # Symbol Content Types must map in Marcel::EXTENSIONS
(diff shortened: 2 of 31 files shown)


rails/app/models/medium.rb
```diff
@@ -0,0 +1,42 @@
+class Medium < ApplicationRecord
```
albertchae, line 1: This is a super pedantic nit that you can completely ignore, but any thoughts on configuring rails inflections so we can name the singular model `Media` as well? In a few places when you loop over multiple `media` instances, you call the block argument `media` as well so it seems like you agree on the singular form in this context. Coincidentally [this stackoverflow post covers custom inflections](https://stackoverflow.com/questions/39585774/pluralizations-and-singularizations-inflections-for-media-medium-in-rails-5) with this exact same use case, ignoring the scaffold part. 

To stay on the pedantic tone, I believe we're using this specific meaning which does have a singular form of "media" https://en.wiktionary.org/wiki/media#Noun_3

albertchae (approved):
Changes look good, new validation gem looks great. Just a couple questions about the new media partial.

rails/app/controllers/dashboard/stories_controller.rb
```diff
@@ -21,9 +21,12 @@ def new
 
     def create
       authorize Story
-      @story = community_stories.new(story_params)
+      @story = community_stories.new(story_params.except(:media))
 
       if @story.save
+        story_params[:media].each do |media|
+          m = @story.media.create(media: media)
+        end
```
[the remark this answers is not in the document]
lauramosher: Normally, yes. Chatting with @rudokemper we are moving to a multiple step process so media upload happens separately. Partially because we want to add some cool media-related features, partially because we’ve gotten some feedback that having them linked can be frustrating.

I could go either way on implementing the transaction for now and removing it in the future, knowing it’s going to change.

rails/app/views/dashboard/stories/_media.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <% else %>
+    <%= image_tag(media, width: size) %>
+  <% end %>
+<% elsif media.video? %>
+  <video
+    controls
+    disablePictureInPicture
+    controlsList="nodownload"
+    id="video-player-<%= media.blob.id %>"
+    poster="<%= url_for(media.preview(resize_to_limit:[nil, size])) if media.previewable? %>"
```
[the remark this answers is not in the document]
lauramosher: The previous implementation was incorrect 😅😅

rails/app/views/dashboard/stories/_form.html.erb
```diff
@@ -21,50 +21,12 @@
...
-            <%= image_tag(media.preview(resize_to_limit:[150, 150])) %>
-            <%= link_to t("dashboard.actions.destroy"), story_delete_media_path(@story, media.signed_id), class: "delete-link", method: :delete, data: {confirm: t("dashboard.actions.confirm"), turbo_confirm: t("dashboard.actions.confirm"), turbo_method: :delete}, remote: true %>
-        <% else %>
-          <%= media.filename %>
-          <%= link_to t("dashboard.actions.destroy"), story_delete_media_path(@story, media.signed_id), class: "delete-link", method: :delete, data: {confirm: t("dashboard.actions.confirm"), turbo_confirm: t("dashboard.actions.confirm"), turbo_method: :delete}, remote: true %>
-        <% end %>
+        <span class="media-with-controls">
+          <%= render partial: "media", locals: {media: media.media, size: 150} %>
+          <%= link_to t("dashboard.actions.destroy"), story_delete_media_path(@story, media.id), class: "delete-link", method: :delete, data: {confirm: t("dashboard.actions.confirm"), turbo_confirm: t("dashboard.actions.confirm"), turbo_method: :delete}, remote: true %>
+        </span>
```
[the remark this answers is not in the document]
lauramosher: Thanks! 🙏

rails/app/models/place.rb
```diff
@@ -6,8 +6,13 @@ class Place < ApplicationRecord
...
   has_one_attached :name_audio
   validates :name, presence: true
-  validates :photo, blob: { content_type: ['image/png', 'image/jpg', 'image/jpeg'] }
-  validates :name_audio, blob: { content_type: ['audio/mpeg', 'audio/wav', 'audio/mp4', 'audio/m4a', 'audio/x-m4a', 'audio/x-aac', 'audio/x-flac'] }
+  validates :photo, content_type: [:png, 'image/jpg', :jpeg], size: { less_than_or_equal_to: 5.megabytes }
+  validates :name_audio,
+    content_type: [
+      :mp3, :aac, :flac, :mp4a, :wav,
+      'audio/wav', 'audio/m4a', 'audio/x-m4a', 'audio/x-aac', 'audio/x-flac',
+    ],
```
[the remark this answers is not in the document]
lauramosher: Sorta. I had high hopes for the mime type symbol use - but it’s wonky. So some look duplicated and some were added to catch some edge cases. I tried not to add anything that required new rendering (like svgs for example).

rails/app/models/medium.rb
lauramosher: Yeah, I waffled on it. Like for a while. You may have noticed the underlying table is “media”

@rudokemper do you have a preference?

lauramosher: By the way - love the pedantic. Keep it up!

rudokemper: Same!

I'd opt for `Media` in singular.

lauramosher: Well, turns out we can have the best of both by just specifying the irregular singularity of media is media.

PR #941, opened 2023-09-01
chore: truncate checkbox labels and increase column number to 3

rudokemper:
Currently, Stories/Places/Speakers checklists in the CMS edit form are showing the full title, which can be unwieldy with long titles:

image

This PR truncates checkbox labels, and increases the total number of columns to 3 in the CMS edit forms. We can definitely do more to improve the form styling, but this is a quick fix for some online communities who are dealing with this.

Result:

image


rudokemper:
Thanks @albertchae that's a great suggestion. I found that I had to modify the generation of checkboxes somewhat to be able to add a title attribute.

image

PR #942, opened 2023-09-01
fix: Add scope to Place & Story Model to filter out null coords (v2)

rudokemper:
This PR adds a scope to the Place and Story Models to filter out any places that have null coordinates (primarily for the front end of TS Main and Explore).

Although we are currently requiring lat/long fields when creating a new Place in the CMS, the Importer currently permits the user to upload Places without lat/longs, hence introducing a bug where Places without lat/long are passed on to React.

We are now adding the stories associated with the communities when they are fetched, not when they are filtered.

In the future we need to rework the Importer as a whole, but that's a whole 'nother epic.

### DIFF
--- a/rails/app/controllers/api/stories_controller.rb
+++ b/rails/app/controllers/api/stories_controller.rb
@@ -2,7 +2,7 @@ module Api
   class StoriesController < BaseController
     def index
       community = Community.where(public: true).find_by!(slug: params[:community_id])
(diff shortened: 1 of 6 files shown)


lauramosher (approved):

lauramosher:
You'll need to ♻️ rebase to resolve the conflicts

rudokemper:
Ugh, gotta change my line ending settings for the repo -_-

Issue #882, opened 2022-11-12
[Map] Use MapLibre for offline environment


rudokemper:
Just scoping out some eventual work to use MapLibre for the offline environment / profile.

~~I created a Ruby gem [maplibre-gl-rails](https://rubygems.org/gems/maplibre-gl-rails) and tested it with Terrastories to make sure that it works, across all areas of the app. (Currently, the map library is piped from Rails to React to serve the map on the front end, as well as the different map views in the Rails CMS.)~~

~~To use `maplibre-gl-rails` instead of `mapbox-gl-rails` for offline maps, we will have to do the following. This is assuming implementation prior to any Rails / React refactor work in the future.~~

* [ ] ~~Add `maplibre-gl-rails` to the Rails Gemfile.~~
  * [ ] ~~Can we selectively include either `mapbox-gl-rails` or `maplibre-gl-rails` in the Gemfile depending on Docker profile? A nice-to-have to keep filesize down.~~
* [ ] In `Map.jsx`: if offline environment is detected, then use MapLibre functions instead of Mapbox functions. Maybe using `useLocalMapServer` or a smarter way per the Rails environment.
* [ ] Implement a similar change across the Rails CMS, specifically in the `show` dashboards for Places and Stories, and the `edit` dashboard for Themes.
* [ ] Import maplibre-gl instead of mapbox-gl for all of the SCSS assets.
* [ ] Deactivate `mapProjection` (since that has not been implemented for MapLibre yet).

PR #943, opened 2023-09-22
Add and use MapLibre in React if running Terrastories in offline mode

rudokemper:
Addresses one chunk of the work described in #882.

This PR uses MapLibre for the React front end map at the /home/ route, if in offline mode.

* Added `maplibre-gl` to package.json.
* In `Map.jsx`, I created a new prop `mapGL` which is set to Maplibre if `useLocalMapServer` is true (with fallback Mapbox), and applied it across the code.
* Imported the Maplibre CSS files in addition to the Mapbox ones.
* Added the MapLibre logo on the bottom right for proper attribution.
* Changed the prop `mapboxStyle` to be named `mapStyle` instead.

Screenshot from 2023-09-22 15-02-33

The next chunk of work will entail using MapLibre in Rails if in offline mode. I tried to do this but it seemed a bit trickier than expected - setting the map in a view like [this one](https://github.com/Terrastories/terrastories/blob/ca7c36a8bb71fc25dbf3fc364fc96afbdd11d5c4/rails/app/views/dashboard/themes/edit.html.erb#L174) to use maplibre instead of mapbox did not work, and I couldn't wrap my head around why mapbox but not maplibre would work if both packages are added and we're otherwise no longer using [mapbox-gl-rails](https://github.com/nbulaj/mapbox-gl-rails) or seeming to do anything special to import mapbox across Rails.

That said, that can be a different PR.

### DIFF
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -4,8 +4,8 @@ import PropTypes from "prop-types";
 import Minimap from "../vendor/mapboxgl-control-minimap.js";
 import Popup from "./Popup";
 
-import mapboxgl from '!mapbox-gl';
 import 'mapbox-gl/dist/mapbox-gl.css';
+import 'maplibre-gl/dist/maplibre-gl.css';
 
 const STORY_POINTS_LAYER_ID = "ts-points-layer";
 const STORY_POINTS_DATA_SOURCE = "ts-points-data";
@@ -13,9 +13,11 @@ const STORY_POINTS_DATA_SOURCE = "ts-points-data";
 export default class Map extends Component {
   constructor(props) {
     super(props);
-    mapboxgl.accessToken = this.props.mapboxAccessToken;
+    const mapGL = this.props.useLocalMapServer ? require('!maplibre-gl') : require('!mapbox-gl');
+    mapGL.accessToken = this.props.useLocalMapServer ? 'pk.ey' : this.props.mapboxAccessToken;
     this.state = {
-      activePopup: null
+      activePopup: null,
+      mapGL: mapGL
     };
   }
 
@@ -24,7 +26,7 @@ export default class Map extends Component {
     points: PropTypes.object,
     framedView: PropTypes.object,
     onMapPointClick: PropTypes.func,
-    mapboxStyle: PropTypes.string,
+    mapStyle: PropTypes.string,
     mapboxAccessToken: PropTypes.string,
     mapbox3d: PropTypes.bool,
     mapProjection: PropTypes.string,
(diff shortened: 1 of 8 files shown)


rudokemper:
Follow up PR to take care of the Rails part: https://github.com/Terrastories/terrastories/pull/944

rudokemper:
As of today, this PR is pulling in both mapbox and maplibre which will bloat our assets, and we will need to find a way to dynamically determine using only one of the two libraries as per the offline environment or a different means.

One possibility might be to incorporate these mapbox/maplibre React components: https://visgl.github.io/react-map-gl/docs

rudokemper:
I implemented dynamic importing to ensure only one library is being pulled on the React side.

I tried to include the CSS imports as well, but that resulted in things like the map controls or minimap (for Mapbox) not rendering at all.

In these screenshots you can see that the non-utilized map library were not pulled by Webpack:

Mapbox:
image

Maplibre:
image

(Also, small but meaningful thing: I had the Minimap use the same style as the map itself.)

rudokemper:
Would be great to get this one in the new `offline-field-kit` build :)

rails/app/javascript/components/Map.jsx
```diff
@@ -34,15 +34,64 @@ export default class Map extends Component {
...
+    } else {
+      Promise.all([
+        import('!mapbox-gl'),
+        import('../vendor/mapboxgl-control-minimap.js')
+      ]).then(([mapboxGLModule, minimapModule]) => {
+        this.Minimap = minimapModule.default;
+        this.initializeMap(mapboxGLModule.default, false);
+      });
+    }
+  }
```
lauramosher, line 50: I'm wondering if we could move the conditional imports into the constructor rather than on mount. We likely want the libraries imported and ready to go before the component has mounted and our rendering starts. I have an unsubstantiated concern that we may inadvertently re-import these libraries if React decides it needs to remount.

rails/app/javascript/components/Map.jsx
```diff
@@ -320,9 +350,12 @@ export default class Map extends Component {
...
     const homeButton = this.createHomeButton();
-    const navControl = document.getElementsByClassName(
-      "mapboxgl-ctrl-zoom-in"
-    )[0];
+    let navControl;
+    if (this.state.isMapLibre) {
+      navControl = document.getElementsByClassName("maplibregl-ctrl-zoom-in")[0];
+    } else {
+      navControl = document.getElementsByClassName("mapboxgl-ctrl-zoom-in")[0];
+    }
```
lauramosher, line 358: I think we can use `querySelector` or `querySelectorAll` to the same effect, while using pseudo regex matchers so we can avoid the conditional:

```suggestion
    navControl = document.querySelectorAll("button[class$="-ctrl-zoom-in]")[0];
```

lauramosher (commented):
This looks good overall. I have a couple of questions/clarifications on the import placement and usage of the map library switching.

rails/app/javascript/components/Map.jsx
rudokemper: So, I couldn't quite figure this out -- to the best of my knowledge and research, dynamic imports are inherently asynchronous, whereas constructors are not and can't handle Promises. Do you have any suggestions for how to handle this via the constructor?

However, I did implement a check to see if the modules have already been imported by setting that in the state. If they haven't, we import and store the modules, and if not, we skip and call `initializeMap` with the already imported modules.

lauramosher (approved):
👍

PR #944, opened 2023-09-23
Use MapLibre across Rails for offline environment

rudokemper:
Addresses the other chunk of the work described in https://github.com/Terrastories/terrastories/issues/882 (along with #943).

This PR implements MapLibre for the different Rails dashboard views that use a map, and also renames `mapbox_style` to `map_style` across the app.

* Added `maplibre-gl` to package.json (also done in #943 but added here for completeness and testing purposes)
* Import `maplibre-gl` in `dashboard.js` to make it available in Rails.
* For Places edit form, Places show view, Stories show view, and Theme edit form, conditionally use either Mapbox or Maplibre depending on the Rails environment.
* Migration to rename `mapbox_style_url` to `map_style_url`.
* Rename `mapbox_style` to `map_style` and `mapbox_style_url` to `map_style_url` where found.

image
_(Map from Tileserver-GL / Maplibre)_


rudokemper:
As of today, this PR is pulling in both mapbox and maplibre libraries which will bloat our assets, and we will need to find a way to dynamically determine using only one of the two libraries as per the offline environment or a different means.

PR #945, opened 2023-09-23
Add Dd logo to sponsors, change .io to .app in seed data

rudokemper:
The tiniest of PRs updating our seed data to use `.app` instead of `.io` for user emails, and setting the default sponsor logo to be that of Digital Democracy since they are now sponsoring our maintenance costs.


albertchae (approved):
nice!

PR #958, opened 2023-09-25
fix: Use turf center() to handle fitBounds on non-Mercator projections

rudokemper:
This PR implements a fix to get around the [open issue on Mapbox GL](https://github.com/mapbox/mapbox-gl-js/issues/11284) of the fitBounds bounding box not working right in non-Mercator projections. The fix is similar to what @lauramosher [implemented for Explore Terrastories](https://github.com/Terrastories/explore-terrastories/commit/8b98900617f789a5eb35c256f51cb07dd91dda80) and eventually we can use that nice Typescriptified code for the main TS app, but this is a fix in the meantime (prompted by seeing this break the app in production).

Small note that I tried to use `featureCollection()` as per the solution in Explore, but it did not render the WSEN bounds from the `framedView` prop in a full GeoJSON format as I expected, so I used a different turf function `bboxPolygon()`.


lauramosher (approved):
> Small note that I tried to use featureCollection() as per the solution in Explore, but it did not render the WSEN bounds from the framedView prop in a full GeoJSON format as I expected, so I used a different turf function bboxPolygon().

This makes sense. The explore code handles bounds by calculating them from the feature collection rather than relying on extra code to generate and store them as a bounding box.

This looks good!

PR #961, opened 2023-09-28
Add login instructions to documentation

julinvictus:
I couldn't find instructions for logging in anywhere in the documentation, adding it to that. I decided to just add a link to `seeds.rb` so we don't need to update documentation if ever changed.
Screen Shot 2023-09-27 at 5 58 40 PM


rudokemper (approved):
Thank you! Just a small recommended fix:

To find usernames and passwords to login to Terrastories, check [seeds.rb](https://github.com/Terrastories/terrastories/blob/master/rails/db/seeds.rb) (the admin username is probably the most useful for most dev purposes).

PR #962, opened 2023-09-29
Allow some basic HTML tags in story description

julinvictus:
Screen Shot 2023-09-28 at 4 01 03 PM
Screen Shot 2023-09-28 at 4 02 35 PM
Screen Shot 2023-09-28 at 6 11 19 PM
Screen Shot 2023-11-01 at 7 26 19 PM

### DIFF
--- a/rails/app/javascript/components/Story.jsx
+++ b/rails/app/javascript/components/Story.jsx
@@ -50,7 +50,7 @@ const Story = props => {
             {story.title}
             {story.permission_level === "restricted" && " 🔒"}
           </h6>
-          <p className="description">{story.desc}</p>
+          <p className="description" dangerouslySetInnerHTML={{ __html: story.desc }}></p>
           {
             story.media &&
             story.media.map(file => (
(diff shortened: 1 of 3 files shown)


rudokemper:
Nice work, @julinvictus!

Can you make this work on the React side as well?

rudokemper (approved):
Looks good! We'll just need to make sure this also works in the React front end. I think it'd be fine to merge this  as-is and do a follow-up PR, or amend this PR to also get the HTML tag working in React.

julinvictus:
> Looks good! We'll just need to make sure this also works in the React front end. I think it'd be fine to accept this as-is, but to merge we should first do a follow-up PR, or amend this PR to also get the HTML tag working in React.

I can take a look at the react part

julinvictus:
When you say React you mean this part here, right?

julinvictus:
When you say React you mean this part here, right?
Screen Shot 2023-10-16 at 7 47 31 PM

rudokemper:
Correct

PR #963, opened 2023-09-29
Remove leading/trailing whitespace for imported has_many fields

rudokemper:
This PR solves a pain point that a number of users are currently having with the importer, having to do with leading whitespaces in the `stories` and `places` fields for Speaker CSV import, which are used to set up a has_many relationship with rows in the Story and Place models (find or create).

These users are trying to import CSVs that have spaces between the values:

image

Currently, the importer will split these up and create them as follows (note the leading whitespace):

```
# => ["Chris /"Kelisi/" Flink", " Frederick /"oom Bree/" Valentijn"]
# => ["Kwango", " Bakaafeti", " Kogondia", " Shatumaipa"]
```

This results in the importer creating duplicate places ` Bakaafeti`, ` Kogondia`, ` Shatumaipa`, and duplicate speaker ` Frederick "oom Bree" Valentijn` because they are distinct from the places / speakers that already exist, but without the leading whitespace. (They are not actually dupes of course because the string is unique, but they appear that way to the user.)

I think for most of our users, it makes sense to have a space for legibility purposes (or whatever tools they are using to format their CSVs end up inserting one). Rather than forcing them to remove all spaces before importing, I think it's good to implement this fix to strip all leading/trailing whitespaces for these fields. (I think it's not as much of an issue for the other fields that don't explicitly set up a relationship with rows from another model.)

(I _think_ the pre-2022 refactor importer also removed trailing/leading whitespaces, so some long-time users are finding that their CSVs which used to work are now having this issue.)

I tested this out with a Stories CSV that was creating >50 "duplicate" places and ~15 "duplicate" stories; this fix ensured all of them were correctly setting up has_many relationships with the existing whitespace-free places / stories that were already imported.

### DIFF
--- a/rails/app/models/concerns/importable.rb
+++ b/rails/app/models/concerns/importable.rb
@@ -120,7 +120,7 @@ module Importable
 
         # Find or create has_many* relationships
         @klass.associated_attribute_names.each do |association|
-          values = attributes[association].to_s.split(",")
+          values = attributes[association].to_s.split(",").map(&:strip)
           attributes[association] = values.map do |name|
             association.to_s.singularize.classify.constantize.find_or_create_by(name: name, community_id: @community_id)
           end
--- a/rails/app/models/concerns/importable.rb
+++ b/rails/app/models/concerns/importable.rb
@@ -120,9 +120,9 @@ module Importable
 
         # Find or create has_many* relationships
         @klass.associated_attribute_names.each do |association|
-          values = attributes[association].to_s.split(",").map(&:strip)
+          values = attributes[association].to_s.split(",")
           attributes[association] = values.map do |name|
-            association.to_s.singularize.classify.constantize.find_or_create_by(name: name, community_id: @community_id)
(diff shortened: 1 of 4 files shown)


rails/app/models/concerns/importable.rb
```diff
@@ -120,7 +120,7 @@ def save_importable_records
 
         # Find or create has_many* relationships
         @klass.associated_attribute_names.each do |association|
-          values = attributes[association].to_s.split(",")
+          values = attributes[association].to_s.split(",").map(&:strip)
```
lauramosher, line 123: Rather than mapping here, I think we can do it when we loop on line 125 `create__by(name: name.strip, …)`. It will save us from doing the same map loop twice and it will set a precedent for any other data manipulation we might need to do in the future.

lauramosher (commented):
Can we add a test for this use case? I believe I added importer tests on the Story/Place/Speaker models that we can extend for this.

lauramosher (approved):

PR #966, opened 2023-10-02
fix: on React, load marker images before adding layers

### LINKED ISSUE #965: [Bug][React] Race condition issue: map marker layers added to map before image is loaded
We are loading map marker images asynchronously with `map.loadImage` and sometimes the layer is added before the image load finishes, resulting in a race condition issue where the layer renders on the map without images for the first set of markers:

image

If you zoom in and out, the images will load for any new markers that appear, 

image

but the original set from the first map extent and zoom will not appear.

Let's figure out an optimal way to ensure that the layers are only added to the map once the images have loaded.

rudokemper:
Resolves #965.

In the Map component, this PR refactors the `this.map.on("load")` event to prioritize loading the marker images before adding the Terrastories point data and layers. This approach addresses a race condition where the layers might sometimes be added to the map before the marker images had fully loaded.

To improve code clarity, I've also extracted the layer addition logic into its own function, `addPlaceMarkerLayers()`.

Before:

image

After:
image

In so doing, the following warnings in the console have also gone away:
```
mapbox-gl.js:31 Image "ts-marker" could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.
mapbox-gl.js:31 Image "ts-marker-cluster" could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.
```

### DIFF
--- a/rails/app/javascript/components/Map.jsx
+++ b/rails/app/javascript/components/Map.jsx
@@ -49,63 +49,23 @@ export default class Map extends Component {
     });
 
     this.map.on("load", () => {
-
-      // Add map point data to the map
-      this.addMapPoints();
-
-      // Add mapbox markers to the map
-      this.map.addLayer({
-        id: STORY_POINTS_LAYER_ID,
-        source: STORY_POINTS_DATA_SOURCE,
-        filter: ['!', ['has', 'point_count']], // single point, non-cluster
-        type: "symbol",
-        layout: {
-          "icon-image": "ts-marker",
-          "icon-padding": 0,
-          "icon-allow-overlap": true,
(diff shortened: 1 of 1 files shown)


rudokemper:
@lauramosher I know our goal is to eventually switch over to the Explore UI which does not have this issue, but hoping this could work as a temporary resolution. This race condition issue shows up every time for some of our online users and it's not great when they're doing a demo.

lauramosher (approved):

Issue #948, opened 2023-09-23
[Rails] Let's improve our title tags across the app


rudokemper:
Currently, the title tags for Terrastories aren't the greatest. Both the welcome screen and the map view are just `Terrastories`. Let's improve this. See acceptance criteria below.

## Acceptance criteria:

- [ ] Welcome screen without having logged in is `Welcome - Terrastories`
- [ ] Welcome screen when logged in is `Welcome - Terrastories: [community name]`
- [ ] Map view is `Terrastories: [community name]`

ice1080:
Just opened up a PR for this one: https://github.com/Terrastories/terrastories/pull/967
Let me know if I need to update anything!

PR #967, opened 2023-10-03
issue 948 - title tags update

ice1080:
For issue #948 
Open to suggestions on better ways to do this. I'm fairly new to ruby.

### DIFF
--- a/rails/app/views/layouts/application.html.erb
+++ b/rails/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Welcome - Terrastories<%= @community&.name ? ": #{@community.name}": '' %></title>
+    <title><%= t("welcome") %> - Terrastories<%= @community&.name ? ": #{@community.name}": '' %></title>
     <%= favicon_link_tag asset_path('favicon.ico'), :rel => 'icon', :type =>  'image/png' %>
     <%= csrf_meta_tags %>
     <%= csp_meta_tag %>
--- a/rails/app/views/layouts/application.html.erb
+++ b/rails/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Welcome - Terrastories<%= @community ? ': ' + @community.name : '' %></title>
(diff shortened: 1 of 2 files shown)


rails/app/views/layouts/react.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Terrastories</title>
+    <title>Terrastories - <%= @community.name %></title>
```
rudokemper, line 4: Set this to `Terrastories: <%= @community.name %></title>`

rudokemper (changes requested):
Just one small fix! Thank you.

rails/app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>Terrastories</title>
+    <title>Welcome - Terrastories<%= @community ? ': ' + @community.name : '' %></title>
```
lauramosher, line 4: Let's use interpolation rather than string concatenation and check for community name in addition to community existence:

```suggestion
    <title>Welcome - Terrastories<%= @community&.name ? ": #{@community.name}": '' %></title>
```

In particular, this will ensure that if community exists but name is somehow nil, it won't crash the page.

```shell
irb> ': ' + nil
TypeError (no implicit conversion of nil into String)
irb> ": #{nil}"
=> ": "
```

@rudokemper Do we also want to localize "Welcome" ?

lauramosher (changes requested):

rails/app/views/layouts/application.html.erb
rudokemper: Great suggestions @lauramosher and yes - that would be a welcome :facepalm: change.

rails/app/views/layouts/react.html.erb
ice1080: Pushed this update!

rails/app/views/layouts/application.html.erb
ice1080: Pushed the nil safe and interpolation update, as well as localization!

lauramosher (approved):

Issue #956, opened 2023-09-23
[Rails] Don't pass orphaned Stories (e.g. no Places or Speakers) to React


rudokemper:
Currently, it's not possible to create a Story without selecting at least one Place and at least one Speaker.

However, it is possible to delete Places and Speakers, which if done can leave a Story "orphaned" (e.g. it exists but has no Places, or Speakers, or both).

It's fine to keep it this way on the Rails side for now -- we need to think about our data structures more -- but we should handle this on the React side.

What is happening now is that the stories will still show up in StoryList. Only, if there are no Speakers, nothing will show at the top, and (worse) if there are no Places, clicking on the card does nothing.

Let's filter out any stories that have 0 Places or 0 Speakers on the React front end. This also means we should filter out any markers on the map that ONLY have stories that have been filtered out.

## Acceptance criteria:

- [ ] If I create a Story, a Place, and a Speaker, and then I delete the Place OR the Speaker (or both), the Story should continue to exist in Rails but it should not be passed to React (e.g. it doesn't show up in StoryList component).
- [ ] If I create a Story, a Place, and a Speaker, and I delete the Speaker, the Story should not be passed to React, and the Place should also not be passed to React (IF is the only Story associated with it; it should continue to appear on the map if it has other Stories with the minimum validation of >1 Place and >1 Speaker).

ice1080:
I can take this one!

ice1080:
Question on this one, assuming for @rudokemper, I'll be filtering out the orphaned stories from dashboard/stories_controller, but should do the same to api/stories_controller? I'm less familiar with what the api module is used for

ice1080:
Thanks for the follow up! 
Regarding this issue, I've got the changes locally but wanted to make sure there are no tests around the pages or dashboard controllers (except Manage Community). Is this correct or am I not seeing them?

A related note, here is a potential way to avoid orphaned parents in this situation. Our use case would be slightly more complex since there are two children to check in the parent, but could still work. 
https://stackoverflow.com/questions/5866378/delete-orphaned-parent

rudokemper:
Thank you! I don't believe we have any tests for the pages or controllers. That would be a welcome addition, but I'm also happy to test your work locally for this.

[...]

PR #971, opened 2023-10-06
issue 956 - exclude orphaned stories

ice1080:
For issue https://github.com/Terrastories/terrastories/issues/956

### DIFF
--- a/rails/spec/requests/api/public_stories_spec.rb
+++ b/rails/spec/requests/api/public_stories_spec.rb
@@ -29,10 +29,10 @@ RSpec.describe "Public Stories Endpoint", type: :request do
   context "filters and sort" do
     let!(:place_1) { create(:place, community: community, region: "the internet") }
     let!(:place_2) { create(:place, community: community, type_of_place: "online") }
-    let!(:place_3) { create(:place, community: orphan_community, type_of_place: "being deleted") }
+    let!(:place_being_deleted) { create(:place, community: orphan_community, type_of_place: "being deleted") }
     let!(:speaker_1) { create(:speaker, community: community) }
     let!(:speaker_2) { create(:speaker, community: community, speaker_community: "ruby for good") }
-    let!(:speaker_3) { create(:speaker, community: orphan_community) }
+    let!(:speaker_being_deleted) { create(:speaker, community: orphan_community) }
 
     # Story:
     # - place 2
@@ -88,33 +88,31 @@ RSpec.describe "Public Stories Endpoint", type: :request do
 
     # Story:
     # - place 1
-    # - type of place: being deleted
-    # - speaker 3
+    # - speaker - will be deleted in test
     # - language: English
-    let!(:story_4) do
+    let!(:story_with_no_speaker) do
       create(
         :story,
         community: orphan_community,
         topic: "test work",
         language: "English",
         places: [place_1],
-        speakers: [speaker_3],
+        speakers: [speaker_being_deleted],
         permission_level: :anonymous
       )
     end
 
     # Story:
-    # - place 3
-    # - type of place: being deleted
+    # - place - will be deleted in test
     # - speaker 1
     # - language: English
-    let!(:story_5) do
+    let!(:story_with_no_place) do
       create(
         :story,
         community: orphan_community,
         topic: "test work",
         language: "English",
-        places: [place_3],
+        places: [place_being_deleted],
         speakers: [speaker_1],
         permission_level: :anonymous
       )
@@ -219,8 +217,10 @@ RSpec.describe "Public Stories Endpoint", type: :request do
     end
 
     it "orphaned stories are excluded" do
-      speaker_3.destroy
-      place_3.destroy
+       # ensures that story_with_no_speaker does not show up in list:
+      speaker_being_deleted.destroy
+      # ensures that story_with_no_place does not show up in list:
+      place_being_deleted.destroy
 
       get "/api/communities/orphaned_community/stories"
       expect(json_response["total"]).to eq(0)
--- a/rails/spec/requests/api/public_stories_spec.rb
+++ b/rails/spec/requests/api/public_stories_spec.rb
@@ -2,7 +2,6 @@ require "rails_helper"
 
 RSpec.describe "Public Stories Endpoint", type: :request do
   let!(:community) { FactoryBot.create(:public_community, name: "Cool Community") }
-  let!(:orphan_community) { FactoryBot.create(:public_community, name: "Orphaned Community") }
 
   it "returns 404 when community can't be found" do
     get "/api/communities/unknown/stories"
--- a/rails/spec/requests/api/public_stories_spec.rb
+++ b/rails/spec/requests/api/public_stories_spec.rb
@@ -29,10 +29,8 @@ RSpec.describe "Public Stories Endpoint", type: :request do
   context "filters and sort" do
     let!(:place_1) { create(:place, community: community, region: "the internet") }
     let!(:place_2) { create(:place, community: community, type_of_place: "online") }
-    let!(:place_being_deleted) { create(:place, community: orphan_community, type_of_place: "being deleted") }
     let!(:speaker_1) { create(:speaker, community: community) }
     let!(:speaker_2) { create(:speaker, community: community, speaker_community: "ruby for good") }
-    let!(:speaker_being_deleted) { create(:speaker, community: orphan_community) }
 
     # Story:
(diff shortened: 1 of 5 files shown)


rails/spec/requests/api/public_stories_spec.rb
```diff
@@ -83,6 +86,40 @@ def json_response
       )
     end
 
+    # Story:
+    # - place 1
+    # - type of place: being deleted
```
rudokemper, line 91: I would take `type of place` out of the comments here and in `story_5` since it's a different field and not relevant to the tests.

rails/spec/requests/api/public_stories_spec.rb
```diff
@@ -127,5 +164,14 @@ def json_response
...
     end
+
+    it "orphaned stories are excluded" do
+      speaker_3.destroy
+      place_3.destroy
+
+      get "/api/communities/orphaned_community/stories"
+      expect(json_response["total"]).to eq(0)
+      expect(json_response["stories"].length).to eq(0)
+    end
```
rudokemper, line 175: Could be worthwhile to also test for a Story to not show if it has 0 Speakers but >1 Places, and 0 Places but >1 Speakers.

rudokemper (changes requested):
Thanks for this! Left a few quick comments on the tests.

rails/spec/requests/api/public_stories_spec.rb
ice1080: Hey @rudokemper! This is actually what I did with how the stories are setup, but the numbered speaker and place don't make it very obvious, I'll update the variable names for clarity.

ice1080:
> Thanks for this! Left a few quick comments on the tests.

Thanks for the feedback @rudokemper. Pushed updates for both of these!

julinvictus:
@ice1080 You think you can rebase your branch so we dont have the conflicts?

ice1080:
@julinvictus rebase should now be complete :+1:

julinvictus (approved):

julinvictus (dismissed):

rails/spec/requests/api/public_stories_spec.rb
```diff
@@ -180,5 +215,16 @@
...
+    it "orphaned stories are excluded" do
+       # ensures that story_with_no_speaker does not show up in list:
+      speaker_being_deleted.destroy
+      # ensures that story_with_no_place does not show up in list:
+      place_being_deleted.destroy
+
+      get "/api/communities/orphaned_community/stories"
+      expect(json_response["total"]).to eq(0)
+      expect(json_response["stories"].length).to eq(0)
+    end
```
lauramosher, line 228: I'd like to verify if this test is necessary as I believe there are specs relating to this already (see lines 165/175 in this file). If you remove the code in the `api/stories_controller.rb` and rerun this test (after adjusting the `json_response` -> `json_meta` for `total`, does it still pass? If so, let's remove it.

rails/spec/requests/api/public_stories_spec.rb
```diff
@@ -83,6 +86,38 @@
...
+      create(
+        :story,
+        community: orphan_community,
+        topic: "test work",
+        language: "English",
+        places: [place_being_deleted],
+        speakers: [speaker_1],
+        permission_level: :anonymous
+      )
+    end
```
lauramosher, line 119: I'd rather utilize the existing story/place/speaker setup rather than adding new ones. I know it makes it easier to conceptually read "place_being_deleted" and the like, but it does slow down the tests considerably since we are forcefully creating and recreating all of these stories, places, and speakers for each test.

rails/spec/requests/home_request_spec.rb
```diff
@@ -20,5 +20,51 @@
...
+          places: [place_2],
+          speakers: [speaker_2],
+          desc: "This story should not be visible because it has no place",
+          permission_level: :anonymous
+        )
+      end
+
+      def json_response
+        JSON.parse(response.body)
+      end
```
lauramosher, line 54: I believe `json_response` (and `json_meta`/`json_body` are all available from `spec/support/api_helpers.rb` now. I also don't see this used in your spec, so either way, we can remove it. :)

rails/spec/requests/api/public_stories_spec.rb
```diff
@@ -180,5 +215,16 @@
...
     end
+
+    it "orphaned stories are excluded" do
+       # ensures that story_with_no_speaker does not show up in list:
+      speaker_being_deleted.destroy
+      # ensures that story_with_no_place does not show up in list:
+      place_being_deleted.destroy
+
+      get "/api/communities/orphaned_community/stories"
+      expect(json_response["total"]).to eq(0)
```
lauramosher, line 226: The Stories API does not return `total` under `json_response` anymore, it's available under `json_meta`:

```suggestion
      expect(json_meta["total"]).to eq(0)
```

rails/spec/requests/home_request_spec.rb
```diff
@@ -20,5 +20,51 @@
...
+        login_as user
+      end
+
+      it "excludes stories with no speaker and/or place" do
+        speaker_1.destroy
+        place_2.destroy
+        get "/home"
+
+        expect(response.body).not_to include('This story should not be visible because it has no speaker')
+        expect(response.body).not_to include('This story should not be visible because it has no place')
```
lauramosher, line 66: This works; however, if you wanted to make use of the `json_response` helper, we can set this test up similar to the one you had in the api spec:

```ruby
expect(json_response["stories"].length).to eq(0)
```

lauramosher (changes requested):

lauramosher:
My bad on submitting that without a comment. @ice1080 — thank you so much for your contribution!

I apologize for the mixed signals here as there have been a lot of changes going on that have changed the landscape of the work you have been doing for us. This looks great and the code fix for the home rendering of the map is a great help!

ice1080:
I'll get to work on these changes @lauramosher. Looks like someone did half the work of this story in some unrelated issue(s), which is why half my work and tests are no longer valid, which are most of your comments, since I created this PR over a month ago.

lauramosher:
@ice1080 I know. I am so sorry for the late review.

rails/spec/requests/home_request_spec.rb
ice1080: Ya this one was added to api_helpers after my PR was created. Removing from file

rails/spec/requests/api/public_stories_spec.rb
ice1080: Another unfortunate case of code being changed after the PR was created. Looks like the tests in lines 165/175 were created by you about a week after my PR was opened. I'll delete this test and the speaker and place being deleted.
For reference, I had originally used some of the old stories, but the first request for change had me create new ones.

rails/spec/requests/home_request_spec.rb
ice1080: Hey @lauramosher, this test is actually not an api test, so the json_response doesn't parse the body since it's html instead of json. Let me know what you'd prefer, Rudo had asked me to create that test so that you had tests on the page contents instead of just the api tests.

ice1080:
@lauramosher pushed my changes. Please see my comments in the above conversations and let me know if you'd like anything else changed. Some of the conversation around this with Rudo was in the issue itself - #956

rails/spec/requests/api/public_stories_spec.rb
lauramosher: Ah, my apologies for the run around.

rails/spec/requests/home_request_spec.rb
lauramosher: Ah, I see where I misunderstood. I noted that the `json_response` helper was included in this file and assumed we were testing a JSON endpoint.

Rudo is correct in that we need a test for this request as it's the one that had the issue. I believe Rudo was ensuring that we tested both the public API and the internal Home map story results since they are powered by two separate controllers.

The home request is powered by a partial that renders JSON (_home.json.jbuilder) and is completely separate from the public API endpoints available in /api/.

Like I said, this works, so I'm not super worried about changing it. If you're interested/wanted to, we could use rspec's `type: :view` attribute and test that this request renders the _home partial and we would then be able to assert on the JSON instead of grepping through HTML blobs.

lauramosher:
@ice1080 Thank you so much!

Do you mind doing another rebase to clean up some of the unrelated changes in this PR?

PR #974, opened 2023-10-10
feature(explore): media preview

lauramosher:
In an effort to streamline the Explore story list view, we want to provide a preview image of a representable media to show that there is media to explore in an expanded view.

This adds the first findable representable media, generates the preview/representation, and then returns the processed URL through the API endpoint.


rudokemper (approved):
:heart_eyes:

PR #982, opened 2023-11-06
Make running tests easier

lauramosher:
This creates a compose test service that adds the ability to easily run specs without needing to understand the intricacies of docker.

The service defines a special entry point so booting the service will automatically run specs. It also allows the standard spec command line interface to work, e.g. --flags and specific test files.


rudokemper (approved):
Love it!

Issue #947, opened 2023-09-23
[Rails] Add the ability for an admin user to upload their own SVG Home icon for the map controls


rudokemper:
Long ago, we customized the standard Mapbox map controls home icon to be a house of the Matawai community:

!https://user-images.githubusercontent.com/31662219/193146179-f0f260d3-b682-4de7-88fe-06ae4466a2aa.png

The intent was always to allow for communities to modify this icon to make it their own, but we haven't gotten around to it until now.

Let's make it possible by allowing an admin user to upload an SVG icon via the CMS dashboard themes page (/member/theme/edit). Then, let's modify the `createHomeButton()` function in `Map.jsx` to set `.home-icon` `background-image` to the community's own SVG icon, or fallback to the existing one in our assets.

This may require some testing of different SVGs to ensure that it will work; there could be some idiosyncrasies to the SVG file structure that prevent it from rendering, in which case we could possibly add some validations.

## Acceptance criteria:

- [ ] A logged in community admin navigates to the Theme edit view in the CMS and sees an option to upload a SVG icon with the label "Custom Home Button", at the bottom the section **Map Settings**.
- [ ] Only a SVG is permitted to be uploaded.
- [ ] The community admin may delete, or replace the SVG.
- [ ] Upon successful upload, the SVG is used for the Home icon on the Map view (/home).
- [ ] The SVG is formatted to fit within the square width/height of the Home icon container.
- [ ] If the SVG is deleted, then the fallback SVG in the assets dir of Terrastories.

LogPRose:
Hello! I am quite new to Rails, I just wanted to ensure that I was editing code in the correct place. for the Map.jsx the dir is /rails/app/javascript/components/Map.jsx/ and for creating the place to upload the SVG icon would be /rails/views/dashboard/themes/edit.html.erb/.
Thanks!

PR #983, opened 2023-12-06
New home button functionality

LogPRose:


rudokemper:
Hi @LogPRose thanks for submitting this!

It looks like you've applied a code formatter that has made quite some changes, which are out of scope for this PR. Can you undo those to keep this focused on just the changes you are submitting?

Also, it would be great if you could add some context in the body of your PR about what you've changed here, and maybe some of the questions you were looking for feedback on in https://github.com/Terrastories/terrastories/issues/947.

PR #984, opened 2023-12-27
[WIP] Use pmtiles with maplibre / streamlined dev & install setup

rudokemper:
Submitting this as a PR so I can leave comments.


rudokemper:
When running `bin/setup` on this branch for the first time (without any pre-existing Docker services or changes to the files), I get some CSP and fetch errors relating to the map in React and the CMS pages that have a map embed:

```
Refused to connect to 'http://rudodev:3000/map/terrastories-map/style.json' because it violates the following Content Security Policy directive: "connect-src 'self' https: mapbox: http://localhost:8080 http://localhost:3000 http://localhost:3035 ws://localhost:3035".
Map.jsx:124 Refused to connect to 'http://rudodev:3000/map/terrastories-map/style.json' because it violates the document's Content Security Policy.
Map.jsx:124 Error: Failed to fetch
    at maplibre-gl.js:31:1
```

bin/install
```diff
@@ -0,0 +1,190 @@
...
+Green='\033[0;32m'
+Purple='\033[0;35m'
+Blue='\033[0;36m'
+
+abort() {
+  printf "%s\n" "$@" >&2
+  exit 1
+}
+
+# Fail fast if tileserver configuration isn't found
```
rudokemper, line 19: Just to be clear, I would say "if map resources are not found in the tileserver directory". Because "tileserver configuration" generally refers to the `config.json` file (which is unique to tileserver), which we are still providing by default.

compose.yaml
```diff
@@ -75,46 +81,39 @@ services:
   localweb:
     profiles:
       - offline
-    extends:
-      file: common-app.yaml
-      service: rails
+    image: terrastories-offline-test
```
rudokemper, line 84: This didn't work for me:

```
 ✘ tileserver Error                                                      0.4s 
 ✘ offline Error                                                         0.4s 
 ✘ localweb Error                                                        0.4s 
Error response from daemon: pull access denied for terrastories-offline-test, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

I think we still need to provide the build context for this method of building terrastories with tileserver-gl, no?

rudokemper: I get some copy errors trying to build `terrastories-offline-test`:

```sh
 => ERROR [prod 4/9] COPY ./rails /api                                                                                                                              0.0s
 => ERROR [builder 2/6] COPY ./rails /api                                                                                                                           0.0s
 => ERROR [builder 4/6] COPY ./map /api/public/map                                                                                                                  0.0s
------
 > [prod 4/9] COPY ./rails /api:
------
------
 > [builder 2/6] COPY ./rails /api:
------
------
 > [builder 4/6] COPY ./map /api/public/map:
------
Dockerfile:54
--------------------
  52 |     
  53 |     # Configure Map Package
  54 | >>> COPY ./map /api/public/map
  55 |     RUN curl -fsSL "https://t.ly/OTZpR" -o "/api/public/map/terrastories-map/tiles.pmtiles"
  56 |     RUN curl -fsSL "https://github.com/protomaps/basemaps-assets/archive/refs/heads/main.tar.gz" | tar xzf - -C /api/public/map/terrastories-map --strip-components=1 basemaps-assets-main/fonts basemaps-assets-main/sprites
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 78505b1b-92e8-467e-a316-99d544772b2e::1xp0b3v6p4ws3u1obo68j56w6: "/map": not found
```

Aside from the weird ./rails -> ./api copy error, it seems we probably shouldn't need to copy anything relating to a default offline pmtiles for the terrastories-with-tileserver flavor.

lauramosher: > `docker build -t terrastories-offline-test rails/.`

This is not quite right anymore.

`docker build -t terrastories-offline-test -f rails/Dockerfile .` The context needs to be from the root folder, not nested within `/rails`.

> it seems we probably shouldn't need to copy anything relating to a default offline pmtiles for the terrastories-with-tileserver flavor.

This is not quite right. The intention is to use the same image for tileserver, protomaps, or mapbox. We always bundle offline-ready pmtiles for the image so it JustWorks™ if you're trying to set it up for the first time on your own with no other map configuration.

rudokemper: That mostly makes sense - it only partially does not insofar as I've only gotten to this step by having tileserver map configuration files to begin with (hence bypassing the abort-if-tileserver-configuration-isn't-found check in the `bin/install` script). 

That said, I'm fine with always bundling a pmtiles file.

bin/install
```diff
@@ -0,0 +1,190 @@
...
+  fi
+fi
+
+echo
+ohai Configuring Terrastories default map package…
+# Download default .pmtiles
+if [ ! -f map/terrastories-map/tiles.pmtiles ]; then
+  curl -fsSL "https://t.ly/OTZpR" -o "map/terrastories-map/tiles.pmtiles" --create-dirs
+fi
+curl -fsSL "https://github.com/protomaps/basemaps-assets/archive/refs/heads/main.tar.gz" | tar xzf - -C map/terrastories-map --strip-components=1 basemaps-assets-main/fonts basemaps-assets-main/sprites
```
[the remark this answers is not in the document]
lauramosher: I purposely included this setup.

1. It's a good middle ground for deprecation and moving to the supported offline setup that uses PM tiles.
2. Just because they are installing with an existing Tileserver configuration doesn't exclude them from using pmtiles. The config allows for swapping between the two.
3. This will become even more relevant as we add additional functionality that will allow us to convert their mbtiles and styles into pmtiles and styles.

compose.yaml
rudokemper: Anyway, having now successfully built `terrastories-offline-test`, and seeing the expected console messages from `bin/serve`,

```sh
[+] Running 6/6
 ✔ Volume "terrastories_postgres_data"  Created                                                                                   0.0s 
 ✔ Volume "terrastories_bundler"        Created                                                                                   0.0s 
 ✔ Container terrastories-tileserver-1  Started                                                                                   0.1s 
 ✔ Container terrastories-db-1          Started                                                                                   0.1s 
 ✔ Container terrastories-localweb-1    Started                                                                                   0.0s 
 ✔ Container terrastories-offline-1     Started                                                                                   0.0s 
You can access Terrastories at: http://rudodev
You can access Tileserver at: http://rudodev/tileserver
```

Running either of these URLS in the browser (or localhost, 127.0.0.1, terrastories.local) yields an Nginx 404 Not Found page.

bin/install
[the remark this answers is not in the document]
rudokemper: I think it's potentially confusing for most of our offline users who I don't anticipate will want to swap between two different map tile formats, but it's fine to keep it. And any nudging towards pmtiles and notification about the eventual deprecation of tileserver / mbtiles is a good thing.

compose.yaml
lauramosher: Ah, I think I understand the confusion. We will no longer be building a docker image using any of the bin/ scripts. This is a huge change from our current standard. Instead, the bin scripts will pull down prebuilt images from Docker hub.

`terrastories-offline-test` is a stopgap image that needs to be locally built in order to test using the setup and install scripts. Normally, it would read `terrastories/terrastories:latest`.

rudokemper: And I might be blanking on how to properly test this branch currently! I have been using the `bin/` scripts..

lauramosher: Can you check the contents of `/etc/hosts` and `.env.offline` and post them here?

rudokemper: `.env.offline` has the following var `HOST_HOSTNAME=rudodev` and my `/etc/hosts` contains the same for 127.0.0.1. I just removed any other localhost hostnames that were defined (like `terrastories.local`) to see if it would make any difference, but the result is the same on my end.

lauramosher: What happens if you try rudodev:3000?

PR #988, opened 2024-01-09
feature: Update to use Maplibre with Protomaps API

lauramosher:
This PR is backwards compatible with usage of mapbox configuration and maplibre with Tileserver.

It further expands functionality to:

- Allow usage of Maplibre across both front end and backend
- Updates Minimap to work with Maplibre
- Adds configuration to allow the usage of protomaps tiles, base maps, and assets; including generation of Style Specification when an api.protomaps.com style is provided
- Ensures Tileserver is configured to serve all fonts so we can utilize fallbacks when a font is not found.

## Testing

- Create a free API key at protomaps.com or ask a Terrastories steward for one
- If you already have .env file, ensure that it does not contain any of the MAPBOX_* ENV vars set (comment them out)
- Set `PROTOMAPS_API_KEY` to the newly created API key, or the one you received from a steward
- Pull down this PR

### Development

run `docker compose --profile dev up`

If you run into issues with dependencies, run:

`docker compose run --rm web yarn`
`docker compose run --rm web bundle`

### Offline

This assumes you already have the database setup.
This assumes you already have tileserver setup.

If you have neither, run `bin/setup` before proceeding.

1. Update `compose.yaml` service `localweb` image tag to `alpha-0.0.1` which contains a prebuilt image from these changes, OR run `docker build -t terrastories-offline-test -f rails/Dockerfile .` and then update the image to 
    ```
    image:  terrastories-offline-test 
    ```
1. Run `docker compose --profile offline up`


rudokemper (changes requested):
This is so, so awesome! 

A few comments:

* Now that we are using Protomaps for online, the title "Online Map Configuration" in Theme makes less sense. Can we make a change in this PR to call it "Mapbox Map Configuration (Optional)"? And rename "Activate 3D Terrain view for online map" to "Activate 3D Terrain view for Mapbox map"?
* The Rails Place show map zooming behavior is a bit wonky for me in both Mapbox and Maplibre. Sometimes it will work at first, and then stop; other times I can only zoom in/out a little bit. I think it might have to do with panning being disabled. I'd say we can just enable it. I don't think it's undesirable to permit the user to pan around; if they get lost, they can just refresh the page.
* I know we are planning docs overhaul, but I think it could be good to at least have some bare minimum indication of how to set things up, like the Protomaps API key env var. I'm happy to contribute this - will push up a change.

rudokemper (approved):

PR #992, opened 2024-04-10
Replace Devise with has_secure_password

lauramosher:
This PR removes Devise. Since we only use database/password authentication right now, don't send emails, and largely discourage registrations, the bulk of what Devise is entirely unused.

As such, Devise has been removed in favor of Rails' `has_secure_password` with the bcrypt gem (which Devise uses under the hood anyway). The new session management for logging in and out can be found in the new `SessionsController` and helpers to determine if the user is logged in a controller concern `sessions/helpers`.

This PR also:
- Removes the Registrations controller and the links to them (which were only on the community search index)
- Updates said community search index to direct users to Explore interface and removes "sign up" links
- Manually updates Devise's trackable interface (see theme.rb) and removes the "remember me" work that didn't work as expected anyway.


rails/config/locales/en/en.yml
```diff
@@ -34,8 +34,9 @@ en:
   built_with_love: Built with ❤️ by
   close: Close
   community_search:
-    title: "Search for a Community to Explore"
-    coming_soon: "This feature is coming soon. For now, please request an invite to a specific community."
+    title: "Explore Terrastories"
+    description: "You can now search and explore publicly available community stories."
+    button_label: "Start Exploring"
```
rudokemper, line 39: Just curious if this is still accessible anywhere, with devise ripped out?

rudokemper (dismissed):

rails/config/locales/en/en.yml
lauramosher: Yes, it is.

PR #994, opened 2024-04-11
chore: Update Ruby 3 and Rails 7

lauramosher:
This PR updates Ruby version to the latest 3.0.X.
It completes the Rails 6.1 upgrade defaults, now loading 6.1 defaults.

And finally upgrades to Rails 7.0.0 and enables as many of those defaults as possible without making too many changes or updates.

### 1. Default to vips (libvips) rather than imagemagick for image processing

We already used the new `image_processing` gem, but our underlying implementation used imagemagick. This replaces the  imagemagick with vips (libvips) and sets the default processor to :vips.

Video processing and PDF processing changes were already completed.

### 2. Better video preview image creation

Video processing previously rendered the first frame as the preview image. This resulted, frequently, in rendering an entirely black or blank image when the video "faded in." Smart processing is now in place to generate a better preview thumbnail from uploaded videos.

### 3. Image Variant processing are now stored in the database

We now will return an already processed variant from the database record rather than needing to recompile. This greatly speeds up requests when the variant has already been processed.

## Things we'll need to do in the Future™

The remainder of the items in the new framework defaults for 7.0.

In particular, the ones related to SHA1 to SHA256 which will require key rotation, among other things. I figure we should get stable on these settings before we try and tackle those.

There is also a note about [schema changes](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#rails-version-is-now-included-in-the-active-record-schema-dump) that we may need to look into before we do any more database changes.

We also need to do a postgresql version upgrade which will take some coordination to ensure we don't break anything on the hosted online instance.

## Testing these changes

To test this out yourself, you'll need to build your own image:

```
docker build -t terrastories/devcore:ruby3.1 -f rails/Dockerfile . --target devcore
```

 and update Compose:

```yaml
  web:
    profiles:
      - dev
    image: terrastories/devcore:ruby3.1
  test:
    profiles: [test]
    image: terrastories/devcore:ruby3.1
```


rails/app/controllers/dashboard/communities_controller.rb
```diff
@@ -8,6 +8,14 @@ def update
       @community = authorize current_user.community
 
       @community.update(community_params)
+
+      # Append new sponsor logos
+      if file_uploads.present?
+        file_uploads[:sponsor_logos].each do |logo|
+          @community.sponsor_logos.attach(io: logo, filename: logo.original_filename)
+        end
+      end
```
rudokemper, line 17: Just curious to know more about the `sponsor_logo` related changes. Did the previous sponsor logo behavior break with the upgrade to Ruby 3 / Rails 7?

rudokemper (approved):
Looks great and was able to build and run on my setup without issue :+1: 
I didn't benchmark this or anything, but everything seems to run faster!

rails/app/controllers/dashboard/communities_controller.rb
lauramosher: Yeah. Previously, we had `config.active_storage.replace_on_assign_to_many` to false. It allowed us to always append uploaded sponsor logos on assigns (aka: `community.update(params)`). The default behavior is intended to be true, and the flag is being deprecated in 7.1.

The recommendation for appending images on `has_many_attached` is to use `.attach`. I left the GREEN/RED spec for testing that behavior in the code if you're curious to see what the difference is with the settings.

PR #1012, opened 2025-11-03
Sync rails from Heroku production snapshot

luandro:
## Summary

Syncs our repository with the exact Rails/React snapshot currently deployed on Heroku
(`terrastories` app).
Heroku `production/main` contains a one-time export commit (`576f922`) that froze Ruby 3.2.4, Rails
7.1.3, Shakapacker 7.x, and the React 18-era dependencies. This PR commits that slug snapshot onto
`origin/master` so we can review and merge the deployed runtime back into version control.

## What changed
- Introduced `.subdir`, `rails/Aptfile`, and other Heroku slug scaffolding files from the deployed
slug.
- Updated `rails/.ruby-version` to `3.2.4`, Gemfile/Ruby dependencies to the deployed set, and
synced `rails/Gemfile.lock`.
- Brought in the corresponding `package.json`, `yarn.lock`, and React/webpack configuration from
Heroku (React 18, Shakapacker 7).
- Added the E2E Cypress specs and spec mocks that ship in the production slug.
- No code authored manually: every file originates from the deployed Heroku slug so we can align
repo source with production.

## Why
- Current `master` still reflects the old 3.0.x / Rails 7.0 stack, while production has already
moved ahead.
- Capturing the slug in git makes the upgrade diff explicit and unblocks forward work (tests, deploy
 scripts, upgrade PRs).
- Once merged, `master` will match what’s live, enabling future upgrades from a consistent baseline.


luandro:
@codex review

